PR #4, opened 2020-08-07
fix: setup jsx pragma for react

### LINKED ISSUE #3: ReferenceError: h is not defined
Hey

Nice library 🤟🏻

I have issue with `react-colorful`. it fails with `h is not defined`. Seems like microbundle compiles react as preact with `jsx-pragma: h`.

You can check it on codesandbox:
https://codesandbox.io/s/intelligent-jennings-jsjw0?file=/src/App.js

jeetiss:
❤️ fix #3

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "lint": "eslint src/**/*.js examples/src/**/*.js",
     "size": "npm run build && size-limit",
     "test": "uvu -r esm tests",
-    "build": "microbundle --entry src/index.js --output dist/index.js --name react-colorful --css-modules",
+    "build": "microbundle --entry src/index.js --output dist/index.js --name react-colorful --css-modules --jsx React.createElement",
     "prepublish": "npm run build",
     "start-demo": "parcel demo/src/index.html --out-dir demo/dist --open",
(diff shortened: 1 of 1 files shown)


RyanChristian4427:
> P.S. Gonna try to add Preact support soon by building a separate bundle. It shouldn't be a big deal.

I figure it's better to talk here than create a new issue.

Preact via the Preact-CLI, which is likely how most users will use Preact, will fully support React-Colorful out of the box with aliasing. Those that don't use the CLI likely already have their own aliasing system set up which will do the exact same thing.

I think you could reasonably say that React-Colorful "supports" Preact right now. "Supports" being in quotes as it's due to the Preact team's work. 

I don't think there's any reason to target the `h` pragma. It will just add complexity without any real gains in users.

omgovich:
Hey @RyanChristian4427. I love your idea!

The way I see it:
We add "Preact support" section (below "TS support") that includes of two parts:
1) it works out of the box if you're using Preact CLI
2) if you're not, read this guide (link to some Preact aliasing guide).

Sounds good @RyanChristian4427?

PR #10, opened 2020-08-13
Demo: set the background color on the whole body

molefrog:
Setting the color on the entire body element will guarantee that the color is applied to out of borders area. Just in case someone wants to scroll the page, it just feels cleaner. 

F8492DE8-A168-4B8F-BF2F-D843E6F01753


omgovich:
Thanks, @molefrog!

PR #12, opened 2020-08-13
Make dynamic favicon that looks like a picker handle

molefrog:
Just got this crazy idea and decided to implement it. 
Screenshot 2020-08-13 at 18 21 13


omgovich:
My God! That's so good, thanks!

I think we should spend some time on performance optimisations. It seems to me that image rendering (with shadow), tag removing and creating on each onChange may take a while on smartphones or/and low budget computers.

omgovich:
@molefrog Check this out.

- Added `useThrottledEffect` with 500 ms interval. Browsers can't update favicon as fast as we update `color` anyway.
- The background (white circle and shadow) is cached as a separate canvas.
- Got rid of `document.getElementById`. Use `ref` to get the node.
- Render a favicon once if the screen width is less than 768.

omgovich:
@molefrog I'm afraid that previous circles' edges will be visible because of antialiasing. So let's keep the current colution.

molefrog:
> @molefrog I'm afraid that previous circles' edges will be visible because of antialiasing. So let's keep the current solution.

Same, ok let's keep it that way.

omgovich:
Deployed: https://omgovich.github.io/react-colorful/

PR #15, opened 2020-08-18
Readme for V2

omgovich:
@molefrog Could you take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,8 @@ The default **react-colorful**'s input/output format is HEX string (like `#fffff
 <details>
   <summary>How to use another color model</summary>
 
+#### Available pickers
+
 | Import path                  | Value example                | Size (gzipped) |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -56,6 +56,32 @@ const YourComponent = () => {
 };
 ```
 
+## Supported color models
+
+The default **react-colorful**'s input/output format is HEX string (like `#ffffff`). In case you need another color model, we provide 5 additional color picker bundles:
```
molefrog:
```suggestion
The default **react-colorful**'s input/output format is a HEX string (like `#ffffff`). In case if you need another color model, we provide 5 additional color picker bundles:
```

omgovich: Thanks! Are you sure about `The default ... input/output format is a HEX string`. Maybe we should use `the` here cause "HEX string" is definite thing?

PR #18, opened 2020-08-20
HexInput component

omgovich:
In case if the user needs to have a text field to type/paste HEX-color, he could import additional `HexInput` component which works well in pair with the picker and costs just 400 bytes:
```js
import ColorPicker from "react-colorful";
import HexInput from 'react-colorful/HexInput';

const YourComponent = () => {
  return (
    <div>
      <ColorPicker color={color} onChange={setColor} />
      <HexInput color={color} onChange={setColor} />
    </div>
  );
};
```

This component doesn't have any default styles. It's just an input with proper behavior. The developers could place and tweak this component any way they want.

BTW I noticed that ~33% of the `HexInput`'s bundle size was the spread operator polyfill, so I used `React.createElement("input", Object.assign({}, props, extraProps))` to fix that. 

### How to test 

`npm run start-demo`

### DIFF
--- a/src/components/HexInput.js
+++ b/src/components/HexInput.js
@@ -33,6 +33,7 @@ const HexInput = (props) => {
 
   // Spread operator replacement to get rid of the polyfill (saves 150 bytes gzipped)
   const inputProps = Object.assign({}, props, {
+    color: null, // do not add `color` attr to `input`-tag
     value,
     maxLength: 6,
     spellCheck: "false", // the element should not be checked for spelling errors
--- a/src/components/HexInput.js
+++ b/src/components/HexInput.js
@@ -1,0 +1,45 @@
+import React, { useState, useEffect, useCallback } from "react";
+import validHex from "../utils/validHex";
+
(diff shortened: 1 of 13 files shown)


omgovich:
@molefrog Could you take a look?

omgovich:
@molefrog I update "Why react-colorful" section with:
```
ships only a minimal amount of manually optimized color conversion algorithms (while most of the popular pickers import entire color manipulation libraries that increase the bundle size by more than 10 KB and make your app slower).
```

But not sure about the inside the brackets. Probably too aggressive)

molefrog:
@omgovich Should be fine!

src/components/HexInput.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    [onChange]
+  );
+
+  // Take the color from props if the last typed color (in local state) is not valid
+  const handleBlur = useCallback(
+    (e) => {
+      if (!validHex(e.target.value)) setValue(escape(color));
+    },
+    [color]
```
molefrog, line 26: Well, that's smart!

PR #23, opened 2020-08-22
Refactor: Convert into TypeScript

RyanChristian4427:
Currently just a first draft of a TS conversion, [mentioned first here](https://github.com/omgovich/react-colorful/pull/22#issuecomment-678535043). Currently still a few issues, some of which just need time, some which will require a few changes.

How would you feel about merging much of the files in `utils/` into a single file, say `conversions.ts`? As of now it's quite tedious to go through the dozen or so files. Named exports would vastly reduce the amount of files making for a much better developer experience IMO.

The build config has a few problems revolving around type generation, not entirely sure how to fix that yet, if possible. Microbundle is really fighting on type definition generation. Adding the `types` field to the `package.json` ensures the main export (`dist`) has generated files and all users can use them. But for some reason that stops all the other exports from generating any types whatsoever. Not sure what's going on there, will have to look into it more.

Have you thought about not splitting up your library into a bunch of separate modules? With modern tree-shaking, I think it could be beneficial to just export one bundle of named components and have users import the ones they want to use. That would result in the following Microbundle output:

```
Build "reactColorful" to dist:
       2.1 kB: index.js.gz
      1.86 kB: index.js.br
      2.04 kB: index.esmodule.js.gz
      1.82 kB: index.esmodule.js.br
      2.11 kB: index.module.js.gz
      1.87 kB: index.module.js.br
      2.17 kB: index.umd.js.gz
      1.93 kB: index.umd.js.br
```

With the following API: 

```
import { HexColorPicker } from 'react-colorful';
```

Bit bigger, but whatever you don't use will be shaken out with modern build tools. 

It would vastly reduce your build process complexity and package size in exchange for some bundle size for the users not using build tools. Just a thought.


RyanChristian4427:
As I mentioned, there are a few issues with the current build system, resulting in Microbundle never generating types for the non-default modules. This very well might be a bug, and I'll raise an issue. Regardless though, I've been thinking about rewriting it in a form that would most closely resemble [Preact's setup](https://github.com/preactjs/preact) for a few reasons.

Using the current system is quite limiting as Microbundle has some options that are only configured via a `package.json` file, and as one is dynamically generated here, they can not be taken advantage of. For example, [build outputs](https://github.com/developit/microbundle#specifying-builds-in-packagejson). Type outputs might be one of these, though again, this might very well be a bug.

Moving to what I suggest not only opens up a couple more Microbundle options but would also allow for targeted builds (rebuilding just `hex`) and reduce build complexity as `build-packages.js` could entirely be deprecated. Not that it is a monster of a build script, but passing off the file system manipulation and the like to Microbundle would reduce surface area for maintenance.

We would simply track `hsv`, `rgbString`, etc. and place a static `package.json` file within them. All built code would be put into `{hsv,rgbString}/dist` rather than just `{hsv,rgbString}`. They can just as easily still point into the `src` directory for their source files so Microbundle won't need any extra help.

Might whip up a quick demo using the current state of master (so non-ts). This PR certainly doesn't need more things being done in it.

src/utils/compare.ts
```diff
@@ -0,0 +1,20 @@
...
+    if (first[prop] !== second[prop]) return false;
+  }
+
+  return true;
+};
+
+export const equalHex = (first: string, second: string): boolean => {
+  if (first.toLowerCase() === second.toLowerCase()) return true;
+
+  return equalColorObjects(hexToRgb(first), hexToRgb(second));
```
RyanChristian4427, line 19: I'm confused as towards why this is here. Surely if the hex strings are not equal converting them to RGB isn't going to change this fact?

Can this just be removed? Or is there an edge case I'm not thinking of?

RyanChristian4427:
I merged the refactored build process into this just as a test to ensure everything was working properly (it is). Build time are also roughly halved with TS as well. Dropped an average build time of 75-90 seconds locally to sub-50 seconds. 

I ended up removing `src/index.ts` as it wasn't offering much utility at this point. Obviously it would get compiled out in the build stage, but TSC will build definition files for any `.ts(x)` file in its scope. This meant we had a definition file that was just defining the import/export of `hex.tsx`. Didn't see the use. 

The only typings I have yet to work out are those in the HOC you suggested. Little bit difficult to get the generics working as they should while keeping them from getting uncontrollable. 

Everything else should be pretty much good to go. I think the `equalHex` function that I commented about above is the only real thing I'd like to maybe touch.

src/components/ColorPicker.tsx
```diff
@@ -1,14 +1,29 @@
...
+import { ColorModel, AnyColor, HSV } from "../types";
+import { equalColorObjects } from "../utils/compare";
+import formatClassName from "../utils/formatClassName";
+
+interface Props {
+  className: string;
+  colorModel: ColorModel<AnyColor>;
+  color: AnyColor;
+  onChange: (newColor: AnyColor) => void;
+}
```
omgovich, line 16: Could we replace that with something like 👇?
```
interface Props extends ColorPickerBaseProps {
  colorModel: ColorModel<AnyColor>;
}
```

src/components/Hue.tsx
```diff
@@ -1,10 +1,17 @@
...
+
+interface Props {
+  hue: number;
+  onChange: (newHue: { h: number }) => void;
+}
 
-const Hue = ({ hue, onChange }) => {
+const Hue = ({ hue, onChange }: Props) => {
   const handleMove = useCallback(
     (interaction) => {
```
omgovich, line 16: Shouldn't we add `interaction: Interaction` here?

src/components/Saturation.tsx
```diff
@@ -1,19 +1,28 @@
...
     (interaction) => {
       // Saturation and brightness always fit into [0, 100] range
       onChange({
+        h: hsv.h,
         s: interaction.left * 100,
         v: 100 - interaction.top * 100,
       });
     },
-    [onChange]
+    [hsv.h, onChange]
```
omgovich, line 25: Could we avoid adding `hsv.h` here somehow?

omgovich: For example just by a simple solution like this:
```
      onChange({
        s: interaction.left * 100,
        v: 100 - interaction.top * 100,
      }: { s: number, v: number });
```

src/components/HexInput.tsx
```diff
@@ -1,10 +1,16 @@
...
+import { validHex } from "../utils/validate";
 
 // Escapes all non-hexadecimal characters including "#"
-const escape = (hex) => hex.replace(/([^0-9A-F]+)/gi, "");
+const escape = (hex: string) => hex.replace(/([^0-9A-F]+)/gi, "");
+
+interface Props {
+  color: string;
+  onChange: (newColor: string) => void;
+}
```
omgovich, line 11: `color` and `onChange` are both optional. Should we add `?` mark?

src/components/Interactive.tsx
```diff
@@ -1,16 +1,29 @@
...
+}
 
-const Interactive = ({ onMove, children }) => {
-  const container = useRef();
+interface Props {
+  onMove: (interaction: Interaction) => void;
+  children: React.ReactNode;
+}
+
+const Interactive = ({ onMove, children }: Props) => {
```
omgovich, line 19: I'm not good at it, but should we use React.FC type everywhere? I mean:
```ts
const Interactive: React.FC<Props> = ({ onMove, children }) => (
```

See: https://github.com/typescript-cheatsheets/react#section-2-getting-started

src/packages/hsl.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import ColorPicker from "../components/ColorPicker";
+import withColorModel from "../hocs/withColorModel";
+import { ColorModel, ColorPickerBaseProps, HSL } from "../types";
+import { equalColorObjects } from "../utils/compare";
+import { hslToHsv, hsvToHsl } from "../utils/convert";
+
+interface Props extends ColorPickerBaseProps<HSL> {
+  color: HSL;
+}
```
omgovich, line 11: Do we need to define `color: HSL;` here if we have `color: T;` in `ColorPickerBaseProps`?

omgovich: Seems to me that we could go with something like:
```js
const HslColorPicker: React.FC<ColorPickerBaseProps<HSL>> = withColorModel(ColorPicker, colorModel);
```

src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export type AnyColor = string | RGB | HSL | HSV;
+
+export interface ColorModel<T> {
+  defaultColor: AnyColor;
+  toHsv: (defaultColor: T) => HSV;
+  fromHsv: (hsv: HSV) => T;
+  equal: (first: T, second: T) => boolean;
+}
```
omgovich, line 29: Shouldn't `defaultColor` also be `T`?

src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  l: number;
+  [key: string]: number;
+}
+
+export interface HSV {
+  h: number;
+  s: number;
+  v: number;
+  [key: string]: number;
+}
```
omgovich, line 20: May I ask what's the purpose of `[key: string]: number;` lines? Seems strange to me because these objects shouldn't have extra keys.

src/packages/hslString.tsx
```diff
@@ -0,0 +1,22 @@
...
+}
+
+const colorModel: ColorModel<string> = {
+  defaultColor: "hsl(0, 0%, 0%)",
+  toHsv: hslStringToHsv,
+  fromHsv: hsvToHslString,
+  equal: equalColorString,
+};
+
+const HslStringColorPicker: React.FC<Partial<Props>> = withColorModel(ColorPicker, colorModel);
```
omgovich, line 20: An idea:
What if we move all src/packages/*.tsx files to their folders?

I mean:
```
src/packages/rgb.tsx → rgb/index.tsx
src/packages/hslString.tsx → hslString/index.tsx
...
src/packages/hex.tsx → src/index.tsx
```

src/types.ts
RyanChristian4427: It's required for your `equalColorObjects` function. TS can't index an object unless you define how it's indexed.

You'll get the following error, though for HSL & HSV too:

```
Type 'RGB' is not assignable to type 'Record<string, number>'.
Index signature is missing in type 'RGB'.
```

src/types.ts
RyanChristian4427: Yes, yes it should. Thanks.

src/types.ts
omgovich: Got it. Thanks for your explanation!

src/packages/hslString.tsx
omgovich: Update:
Read that first: https://github.com/omgovich/react-colorful/pull/25#discussion_r475801723

src/utils/compare.ts
omgovich: react-colorful support HEX3 colors, so it's possible that user will pass a color like `#ABC` and `"#abc".toLowerCase() !== "#aabbcc".toLowerCase())` so I just convert both colors to RGB to compare they RGB representations. This way is smallest in terms of bundle size baucause `hexToRgb` is in the bundle anyway.

src/components/Interactive.tsx
RyanChristian4427: I've been using Preact heavily for the last 6 months or so and haven't touched React since, so I apologize, some of my typings might be a bit awkward.

src/utils/compare.ts
omgovich: Actually HEX is the same as RGB but formatted differently.
Both models include thee [0-255] numbers. RGB store them in decimal, HEX — hexadecimal (00-ff where ff = 255).

RyanChristian4427: Oh, true. Sorry. Forgot about HEX3/HEX6. Makes sense.

Though I wonder if it would be better to avoid conversion to RGB and the iterating over object indexes altogether.

src/components/HexInput.tsx
RyanChristian4427: I'll use `Partial` instead for type re-usability, but yes, thanks, missed that.

src/components/Saturation.tsx
RyanChristian4427: Sorry, absolutely.

src/components/Hue.tsx
RyanChristian4427: In `Hue`'s props?

src/packages/hsl.tsx
RyanChristian4427: Yeah, absolutely. With a `Partial` as the second layer in that generic though.

src/components/Interactive.tsx
RyanChristian4427: The type is inferred. It's not necessary, even in strict mode. Could type it, but it's essentially like doing:

```
const x: string = "hello world";
```

In fact, there's apparently [reasons to avoid `React.FC` altogether.](https://github.com/facebook/create-react-app/pull/8177)

Sorry, I've been playing around in Preact land for the past 6 months and it has entirely replaced React for me. Few differences in typing the two, so I might be out of touch with React specific typings, and now there seems to be a movement against `React.FC` as of recently.

omgovich: Alright.
I'm actually okay with any solution you choose here)

src/utils/compare.ts
omgovich: No problem)
BTW this code also solves the case when one of the colors doesn't have `#`. Like `equalHex('000', '#000)`. 
`hexToRgb` knows everything about hex-strings anatomy =)

RyanChristian4427: Yeah fair. I'll leave it. If there's performance improvements to be had this PR is not the place to do it. Already too much going on.

omgovich: Agree. PR is huge 😀
It's amazing how much you have done.

src/components/HexInput.tsx
omgovich: Could I ask about advantages of `Partial`?
It feels to me that `?` will it more clear for those who trying to understand how everything works. But I'm not a pro in TS and I probably just doesn't know something)

src/utils/compare.ts
RyanChristian4427: Sorry about that. Hard to incrementally move to TS.

omgovich: Don't worry! Actually you're a hero) And I don't know how to thank you) Feel free to invite me to any of your projects — I'll be glad to help. I'm not good in terms TS yet, but I'm in web-dev since 2008 and know a lot of stuff =)

src/components/HexInput.tsx
RyanChristian4427: So `?` is for a specific field, while `Partial<object>` is the same as labeling every field with `?`. 

In general it's just a shorthand, but it's nice for being able to reuse the interface. The component is a pretty good example. Take the following:

```
interface Props {
  x: string;
}

const Component = (props: Partial<Props>) => {
  const { x }: Props = { x: "", ...props};
  ...
}
```

In the above example, the component will accept a partial representation of `Props`. So `x` is completely optional. We can then reuse that interface in the component to describe our de-structured props to ensure we're not forgetting a field while setting default values.

Had we used `?` in `Props` then `x`, even after being destructured, would actually be defined as `string | undefined`.

Using `Partial` gets around having two interfaces; one for the optional props, and one to tell TS that you know your destructured values are not `undefined`.

src/components/ColorPicker.tsx
```diff
@@ -1,14 +1,26 @@
...
+  colorModel: ColorModel<AnyColor>;
+}
 
-const ColorPicker = ({ className, colorModel, color = colorModel.defaultColor, onChange }) => {
+const ColorPicker = ({
+  className,
+  colorModel,
+  color = colorModel.defaultColor,
+  onChange,
+}: Props): JSX.Element => {
```
RyanChristian4427, line 20: @omgovich This will be something I need your feedback/choice on.

A default rule with the TS-ESLint plugin says that functions should have explicit return types when it can not be absolutely inferred. That would lead to this function needing the `JSX.Element` type while other functions, like in `HexInput.tsx`  don't as the usage of `memo` will immediately narrow down the type. 

Some TS style guides say this is a nuisance and at times the TS team itself says best practice is using inference until you absolutely need to define a type (as the compiler can't work it out itself). Can't link directly, but the small paragraph right above [this link](https://www.typescriptlang.org/docs/handbook/generics.html#working-with-generic-type-variables) says "While type argument inference can be a helpful tool to keep code shorter and more readable, you may need to explicitly pass in the type arguments as we did in the previous example when the compiler fails to infer the type, as may happen in more complex examples."

Pretty much just a stylistic choice. Could just continue as-is, and add a `JSX.Element` or two even if they aren't strictly speaking necessary, or could rely on inference more heavily and disable the rule.

RyanChristian4427: Just something I came across. Keeping as-is might just look like an inconsistency, as you spotted with the use of `React.FC`, when in reality it's just there to help the linter along.

omgovich: It seems to me that the code you just commited is fine. We don't need React.FC since we don't use anything it provides (children, defaultProps, etc).

RyanChristian4427: We don't need `React.FC`, but would `JSX.Element`. Just to clarify, `HexInput` is perfectly fine for the linter as:

```
const HexInput = (props: Partial<Props>) => {
```

but `ColorPicker` needs:

```
const ColorPicker = ({ ... }: Props): JSX.Element => {
```

Because while `HexInput` is exported like:

```
export default React.memo(HexInput);
```

`ColorPicker` is as:

```
export default ColorPicker;
```

It's no problem to add the extra typing, certainly harmless enough, but might look inconsistent. Thought I'd explain why as you had noted the inconsistency when I used `React.FC` in some places but not others. This is the same "issue" I guess, as the linter will call for some explicit typings in some places but not others.

omgovich: You know what I just thought: we could move memo to ColorPicker.tsx since we use HOC that passes colorModel as a prop. 

The reason why I wrapped the component with memo in each package was becase I used defaultProps that can't be applied to menoized component.

RyanChristian4427: Well, it's (potentially) going to be an "issue" across your code base even if you manipulate 1 part into matching the rest is my point.

RyanChristian4427:
Only a couple issues remain (from my side) at this point. Firstly, the HOC, which I don't really know how to fix. I avoid them like the plague as I find them totally unreadable and increase complexity with no return value. Totally just copied your suggestion for it originally.

Secondly is the TS types. Honestly not sure how we can fix that. Unfortunately Microbundle will bundle all types under whatever path we give. For example, if our types path is `dist/packages/hex.d.ts`, our output will be:

```
packages/
--components/
----....
--hocs/
----....
--packages/
----...
```

If it's `packages/packages/hex.d.ts`, we get:

```
packages/
--packages/
----components/
------....
----hocs/
------...
----packages/
-------...
```

And so on and so forth. It's actually impossible for us to target deeper down in the tree as we'd need to. If we edit the `types` field then Microbundle will just nest the types down deeper. I don't know if there is a solution to this, besides rewriting every `package.json` after types are generated with the correct path, though this is horribly awkward. 

Microbundle expects a top level `.d.ts` file that we just can't provide. No amount of shuffling files around will get this to happen as far as I am aware.

build-packages.js
```diff
@@ -37,9 +37,9 @@ fs.readdir(entryDirPath, async (e, files) => {
         main: "index.js",
         module: "index.module.js",
         esmodule: "index.esmodule.js",
-        "umd:main": "dist/index.umd.js",
+        "umd:main": "index.umd.js",
         source: `../src/packages/${file}`,
-        types: "../types/index.d.ts",
+        types: "index.d.ts",
```
omgovich, line 42: Hey @RyanChristian4427! How are you?
Do we still have issues with TS bundling or you have solve them?

src/components/ColorPicker.tsx
omgovich: It just seems better to me to write `React.memo` once instead of copying it for each package.
Feels like the right way because it makes the code simpler.

omgovich:
Hey, @RyanChristian4427. Sorry, I'm a little bit busy migrating to a new notebook, but I'm going to take a closer look as soon as I can.

build-packages.js
RyanChristian4427: Types are definitely still messed up, Microbundle doesn't allow to target any types deeper than top level it seems.

omgovich: Any ideas on how to fix that?

If we won't include `types: "index.d.ts",` to package.json by default, it helps?

If it doesn't I could try to rewrite the build-packages.js to simulate command running inside subfolders. I guess that's not a big deal.

RyanChristian4427: We could move everything from `packages` up a level, maybe edit the build script to just target specific files rather than building the queue dynamically. Or rewrite each `package.json` after the fact with the correct paths. 

I opened an issue on the Microbundle repo just to ensure this is the case as nothing I've tried can "un-bury" the types.

RyanChristian4427:
The types issue is still blocking, so I'll have a play around with that, but I think content wise I'm mostly finished? No `any`'s remain, and using `strict` mode so everything that TS can't infer is typed. 

I ended up removing the HOC altogether as it was causing major issues while not actually providing any value. We were just straight passing props through anyways.

omgovich:
Looks good without HOC, thanks!

RyanChristian4427:
I went back and check the sizes, the only component that had any sort of significant size increase from the current state of master was `HexInput`, but that's been corrected now. I think everything should be within 10 bytes of the original, JS version now.

build-packages.js
omgovich: Just found out that node's exec method os also supports cwd option:
That could make Microbundle ignore the root package options and structure:
`exec('git status', {cwd: '/home/ubuntu/distro'})`

RyanChristian4427: Sorry, I'm not following, how would that help?

Microbundle builds the type definitions into whatever directory is specified in `package.json`. It assumes we have an `index.ts` at the top level and will generate the directory structure to make that happen.

omgovich: Sorry, I understood the problem wrong.
The types are wrong because all entry files are placed at the same level as the files they use, but they should be upper (more "root"), right?

omgovich: Let's just move all package entry files to /src then?
```
src/packages/hex.ts -> src/index.ts (or src/dist.ts)

src/packages/rgb.ts -> src/rgb.ts

...
```

RyanChristian4427: Yeah, pretty much. 

It's a bit difficult to describe over text, but if you clone and build locally it becomes pretty clear. 

Microbundle will toss all children of `src` into the top-most directory specified in the types field of` a `packages.json`. So if we have `"types": "dist/index.d.ts"`, our output would be:

```
dist/
--components/
--packages/
--utils/
...
```

If we target lower, like `"types": "dist/packages/hex.d.ts"` we get:

```
dist/
--packages/
----components/
----packages/
----utils
...
```

Quite difficult.

RyanChristian4427: > Let's just move all package entry files to /src then?
> 
> ```
> src/packages/hex.ts -> src/index.ts (or src/dist.ts)
> 
> src/packages/rgb.ts -> src/rgb.ts
> 
> ...
> ```

We certainly could do. Just need to change the build script to be more explicit in we want built then.

I was attempting to see if there was some way we could keep the current structure but if this works for you then great.

omgovich: Yeah, it works) If we'll find a way to keep the current structure, we could do it a a separate PR. Let's go with moving eveything to /src since we know that it solves everything and allows us to finish this task and merge the changes.

omgovich: Thanks again for your patience. It took a while to understand the problem) My bad)

RyanChristian4427: Oh this is quite awkward, I'm sure I didn't explain it well. Trying to type out directory structure in text form is never easy.

RyanChristian4427:
@omgovich I updated the build process, everything seems to be good. 

I did switch it over to use ESModules, as I figure that better matches the repository with the ES6 import syntax and the like. Let me know if you'd rather keep it as CommonJS though, no problem. Thought I'd give it a shot. 

I did have to enable an experimental flag to import from a `.json` file for getting the `peerDependencies` field but I figure that's unlikely to cause issues.

build-packages.mjs
```diff
@@ -0,0 +1,86 @@
...
+// Read available package sources from the dir
+const modulesArray = [
+  "src/hex.tsx",
+  "src/HexInput.ts",
+  "src/hsl.tsx",
+  "src/hslString.tsx",
+  "src/hsv.tsx",
+  "src/rgb.tsx",
+  "src/rgbString.tsx",
+];
```
RyanChristian4427, line 25: I figured this was the best/easiest way to target the packages for now, totally could be missing a better option though.

RyanChristian4427:
Also, something you may notice is that TSC/Microbundle will output types for all files, even those that are not actually public interfaces. Unfortunately there's no way to stop this besides manual pruning. TS can't yet distinguish between external and internal typings.

So in our bundle for say `HexInput` will be the typings for `hex`, as well as `Interaction`. Neither are public, neither are accessible, yet they're shipped out in the bundle. This shouldn't actually matter as no types will ever be shipped to end users, but I digress. 

Maybe this only bothers me though. Don't know. Thought it worth mentioning that, yes, technically it is working as intended.

build-packages.mjs
omgovich: I'm okay with this, thanks. The updated building script looks good! But I'm afraid about migrationg to mjs/ESM here.

It supported only since Node 13.2.0 which isn't stable yet.

RyanChristian4427: Yeah, I forgot how (not) long ago ESM was actually introduced in a reasonable manner. 

Oh well, no problem reverting. Just a build script anyhow.

RyanChristian4427:
@omgovich 

Should be good to go now. Of course still review away if I've missed something.

omgovich:
@molefrog Could you join us here? Fresh eyes would be good)

omgovich:
Works and looks good! 

@RyanChristian4427 Could I ask you to fix conflicts?

In my opinion, this is how our plan should look like:
1. Fix conflicts
2. Deploy to NPM with beta-label — `v2.3.0-beta.1`
3. Install the beta-package to some TS-projects and test how it works in a real environment.
4. Merge with brand to master and release `v2.3.0`

@RyanChristian4427 Sounds good?

RyanChristian4427:
Sorry, conflicts? Not seeing any on my side.

But rather than publish a whole beta version, we could just use npm/yarn link and test that way? That's what I've been doing as I've went along, testing the components/intellisense typings in Preact TS project. 

I should still do another full run through, but I don't think there's a need for a beta?

omgovich:
Guess you're right. We could just use `link` instead.

omgovich:
Oh, sorry. Everything is okay — there are no conflicts.

RyanChristian4427:
I'm wondering if we should add `@types/react` to the peer dependencies list. While many projects do have the `skipLibCheck` in their `tsconfig` files, some do not. You can just as easily ignore peer dependency warnings but this would be a nice reminder for say TS Preact users.

RyanChristian4427:
Also, I know next to nothing about HSL/HSV so while things look a bit weird to me they might not be? Might just be the wheel -> square conversion, but values don't seem to increase/decrease like I'd expect.

RyanChristian4427:
Additionally, how do you feel about maybe doing named exports with the types? For example:

```
export default HslColorPicker;
export { HSL };
```

```
import HslColorPicker, { HSL } from 'react-colorful/hsl';
```

They could otherwise import the type like `import { HSL } from 'react-colorful/dist/types';` or `import { HSL } from 'react-colorful/src/types';` but I'd say those aren't as good of experiences.

molefrog (approved):

omgovich:
@RyanChristian4427 Merged. Thank you again!

P.S. I love the idea of the named color types export!

RyanChristian4427:
@omgovich 

I'm going to actually rethink a bit of our external/internal definitions for the three (RGB, HSL, HSV). I realized exporting those types would make it a bit more obvious that any string key could actually be added and would work fine (i.e., `{ r: 0, g: 0, b: 0, x:0 }`). 

As of now the user has to more or less dig to find out what HSV is strictly defined as, which almost is a benefit as most wouldn't think of adding extra keys to an HSV object. Not the best situation but I'll try to straighten it out. Should just be able to cast to a different object when doing our color object comparisons, and keep otherwise correct types throughout.

RyanChristian4427:
> Regarding:
> 
> > Have you thought about not splitting up your library into a bunch of separate modules?
> 
> That's the thing I want to avoid because tree-shaking doesn't always work properly.
> 
> Sometimes it doesn't work at all. For example, Parcel bundler still doesn't have a stable version of the tree-shaking feature.

I'm still thinking the output should be combined and user can use named exports to get the parts of the library that they want. While multiple separate inputs -> outputs is on Microbundle's roadmap I don't see any related PRs open (or closed) relating to it. Just old issues.

To continue using the current solution would just be a nightmare I think, especially as time goes on. Every extra output that gets added means more duplicated `.d.ts` files, more duplicated `.css` files, etc. Maybe there are ways to prune these but that just moves the complexity into your build script. Not to mention build time, which is quite slow as well. Seeing as how we can't selectively target modules for compilation (at least not without also generating `tsconfig` files as build time to narrow compilation scope) each new file is also going to exponentially increase build times.

The current worst case scenario is if user wants to use only `HexInput` with 0 tree-shaking of any kind, as odd as that'd be. That's still just 1.6kb of potential waste. And while I know, _**just?!**_, but yes, just. Because that is absolutely nothing compared to the average React app (CRA with TS starts at 120kb). It's minimal even to the average Preact app. Of course, in a more realistic scenario, no one would be trying to use the `HexInput` and the `HexInput` alone. It's pretty much just a normal input element. Of the remaining components, the smallest is the `HsvColorPicker`. If a user wanted to use only it, with 0 tree-shaking, they'd see an 850b bundle size increase. Which is even smaller of an issue.

As was said more than a year ago now, [the popular bundlers will all handle tree-shaking](https://github.com/developit/microbundle/issues/50#issuecomment-491237143). If Parcel/another bundler can't work this out I think that's an issue for them, and not something we should really worry about. [Majority of users use Webpack](https://2019.stateofjs.com/other-tools/#build_tools), and it can tree-shake quite well. 

The move to named exports will see:
  - Faster build times
    - Could be important if more people want to help with this library
  - Smaller overall bundles (both in file size & count)
  - Less surface area for maintenance, as the build script/s won't be needed
    - Again, useful if more people want to help
  - No need for irrelevant files (types for say the hex color picker in the hsv's output, or stylesheets in anything but the main output)
  - Slightly less verbose API, as now I can import both `HexColorPicker` and `HexInput` in the same statement
  - No increase to bundle size with all modern (and widely used) build tools

Working with this library more since I originally asked that question only convinces me further that it is the right path. It is such a pain to work with currently due to the combination of build times and junk output. Don't get me wrong, I understand the value and virtue of a small output but trying to save a rather insignificant amount of bytes for a small minority of users just seems silly. The efforts are totally lost extremely quickly. Only noticeable if someone really cares, and if they care that much, they'd probably be using a build tool that can tree-shake.

I thought I'd bring this up again as I was reading through Microbundle's and some other bundlers' documentation trying to find an easy solution but I couldn't help but think back to this.

omgovich:
Hey @RyanChristian4427. I was also thinking about that recently and moving to tree-shakable named exports seems the right decision.

So, I assume it will be our v3 since those are breaking changes.

PR #25, opened 2020-08-24
refactor: Reworked build system

RyanChristian4427:
I looked into doing this as I was having issues wrangling Microbundle into outputting types in #23. The issue is outlined a bit there.

This would not only be a solution to that problem but also has the potential to be an improvement outside of that. Reduces surface area of maintenance and locally I'm seeing builds take half the time.


package.json
```diff
@@ -12,7 +12,14 @@
     "lint": "eslint src/**/*.js demo/src/**/*.js",
     "size": "npm run build && size-limit",
     "test": "jest tests",
-    "build": "node ./build-packages.js",
+    "build": "npm-run-all --parallel build:*",
+    "build:submodule": "microbundle build --css-modules true --jsx React.createElement",
```
RyanChristian4427, line 16: I'm not the biggest fan of labeling this as "submodule" but it allows us to both build the main and use it as a reusable command for the rest so we're not repeating flags. Could separate it out into two commands, but thought I'd present this first.

rgbString/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "react-colorful-rgb-string",
```
RyanChristian4427, line 2: `build-packages` originally would've presented this as `"name": "react-colorful-rgbString"` but this does not follow [NPM's naming guide](https://docs.npmjs.com/files/package.json#name) as capital letters are not allowed. So I changed this into a hyphenated word.

package.json
```diff
@@ -24,27 +31,27 @@
...
       "limit": "2 KB"
     },
     {
-      "path": "hslString/index.module.js",
+      "path": "hslString/dist/index.module.js",
       "limit": "2 KB"
     },
     {
-      "path": "hsv/index.js",
+      "path": "hsv/dist/index.module.js",
```
RyanChristian4427, line 50: I changed this from `index.js` to `index.module.js` as I assumed it was a mistake? Every other path was for `.../index.module.js`. Let me know if it wasn't and I can revert.

omgovich: Good catch!

rgbString/package.json
omgovich: 👍

omgovich:
Works like a charm!

omgovich (approved):

package.json
omgovich: We can call it `build:dist`. IMO that makes it clearer that this command builds `/dist` folder.
What do you think?

omgovich:
Ready to merge and deploy on your command)

package.json
RyanChristian4427: Fully agree, that's a better name.

RyanChristian4427:
Should be ready to merge. I don't think it'll _need_ a deploy though, as nothing is functionally different.

I did realize that I neglected to copy over the `name` flag for the new scripts, so that's been fixed. I just copied it over as you had it, not sure if you'd like a different format with that.

.gitignore
```diff
@@ -14,15 +14,6 @@ node_modules/
...
 dist/
 
-# packages
-/hex
-/hsl
-/hslString
-/hsv
-/rgb
-/rgbString
-/HexInput
```
omgovich, line 24: **Discussion:**

I feel that this update is going to make the top-level repo structure more complex.
And that'll make the barrier to entry way higher.
I just thought that keeping so many folders in repo root isn't the best idea.

It seems to me that we could combine `build-packages.js` solution and the one you did. I mean to create the same package.json, dist subfolders, run the same command etc. I'm going to create a separate PR to try.

RyanChristian4427: Fair enough. Though I still think this will just move the complexity into your tooling instead. It'll also be slower, though depending on what you're used to, might not be important. 

You'll just need to first generate a valid `package.json`, then run the Microbundle in that directory, rather than the other way around.

omgovich:
Going to close this PR. These ideas and concepts helped us to improve building process in #26.
Thanks @RyanChristian4427 ❤️

RyanChristian4427:
#26 is the successor

PR #26, opened 2020-08-24
Building process improvement

omgovich:
Hey, @RyanChristian4427 👋 Check this out.

I took your concept from #25, but keeps packages ignored by Git.

Also, I made the building process parallel by `Promise.all(...)`

### DIFF
--- a/build-packages.js
+++ b/build-packages.js
@@ -2,48 +2,73 @@ const fs = require("fs");
 const path = require("path");
 const del = require("del");
 const util = require("util");
+const { kebabCase, map } = require("lodash");
+const { peerDependencies } = require("./package.json");
 
 // Convert NodeJS methods to promises in order to use them with `await` statement
 const exec = util.promisify(require("child_process").exec);
 const writeFile = util.promisify(fs.writeFile);
+const makeDir = util.promisify(fs.mkdir);
 
 // Read available package sources from the dir
 const entryDirPath = path.join(__dirname, "src/packages");
 
 fs.readdir(entryDirPath, async (e, files) => {
-  console.log(`⚙️ Building ${files.length} packages...`);
-
-  for (let file of files) {
+  // Bundles a package asynchronously
+  const bundlePackage = async (file) => {
     const { name } = path.parse(file);
     const isMainPackage = name === "hex";
-    const filePath = path.join(entryDirPath, file);
     const outputDirPath = path.join(__dirname, isMainPackage ? "dist" : name);
     const manifestPath = path.join(outputDirPath, `package.json`);
     const bundlerPath = path.join(__dirname, "node_modules/.bin/microbundle");
 
+    // Format a package name according to NPM's naming guide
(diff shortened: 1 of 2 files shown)


RyanChristian4427:
Looks good. We might have to inject the `types` field after the fact though. Just noticed my solution, when merged into the TS branch had a pretty big failing because of this. Little bit of awkwardness with Microbundle.

build-packages.js
```diff
@@ -2,48 +2,73 @@ const fs = require("fs");
...
+      await makeDir(outputDirPath);
       await writeFile(manifestPath, manifestCode, "utf8");
     }
-  }
+
+    // Bundler options
+    const args = {
+      name: packageName,
+      cwd: isMainPackage ? __dirname : outputDirPath,
+      output: `${outputDirPath}/index.js`,
```
RyanChristian4427, line 54:
```suggestion
      output: `${outputDirPath}`,
```

I don't think you need the `.js` here. [Microbundle's `output` is for a directory](https://github.com/developit/microbundle/#all-cli-options-)

RyanChristian4427: Whoops, didn't catch you in time. Sorry.

build-packages.js
```diff
@@ -2,48 +2,73 @@ const fs = require("fs");
...
       // Create `package.json`
       var manifestCode = JSON.stringify({
-        name: `react-colorful-${name}`,
+        name: packageName,
         private: true,
         main: "index.js",
         module: "index.module.js",
         esmodule: "index.esmodule.js",
-        types: "../types/index.d.ts"
+        "umd:main": "dist/index.umd.js",
```
RyanChristian4427, line 40: Oh, and this doesn't look correct. Sorry, didn't catch it.

build-packages.js
omgovich: Don't know how it works, but If I remove `index.js` microbundle names the files like that:
image

omgovich: Looks like Microbundle doesn't look at `main: "index.js"` in `package.json`. Hmmm

omgovich: They use a file name in their own examples, so I think it's okay:
https://github.com/developit/microbundle/#building-a-single-bundle-with-a-fixed-output-name

RyanChristian4427: That's really odd. I don't think I've seen it doing that before. 

Though as I usually stick to TS, I guess I've avoided that issue. Nvm then. Don't know how that happens.

build-packages.js
omgovich: Not sure that I got what you mean)

RyanChristian4427: The `dist`. If you try to use the `umd:main` you'll be pointed at a file that doesn't exist. Should be `index.umd.js`.

RyanChristian4427: Microbundle still generates the right file, it's just the path that's incorrect.

omgovich: Oh, I see. Nice catch, thanks!

PR #28, opened 2020-08-26
ci(actions): Adding compressed-size-action

RyanChristian4427:
Compressed Size Action is a GitHub action that reports changes in the compressed file sizes in PRs. [Source Page](https://github.com/preactjs/compressed-size-action). This will allow you to see, at a glance, whether a PR increases or decreases your bundle size, and by how much. Incredibly useful for a library watching its size like this one. It would catch [regressions like this](https://github.com/omgovich/react-colorful/pull/23#issuecomment-680262792) that would normally require manually checking/comparing of numbers.

As of now, this only runs on a PR, but it is also in the works to support push events. 

Having ran a [quick test here](https://github.com/RyanChristian4427/react-colorful/pull/1) it seems that this may only pick up on changes in a `dist` directory. Will figure out the root of that issue but that can/should(?) be fixed in another PR anyhow. So for now this will only work on the main output (hex).

I don't want to directly link as that will ping them, but pull request #2701 over at [Preact](https://github.com/preactjs/preact) shows exactly what we'll be seeing and I think that will be super applicable for this library. It's an open PR, just a couple down from the top of the list. Sorry if there's a way to link without pinging them but I don't know how to do that. 

Edit: Sure enough, moving `rgb/index.js` and the lot to `rgb/dist/index.js` works like a charm.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,14 @@
+name: Compressed Size
+
+on: [pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: preactjs/compressed-size-action@v2
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,4 +12,4 @@ jobs:
       - uses: preactjs/compressed-size-action@v2
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
-          pattern: "**/dist/**/{index.module.js,index.css}"
+          pattern: "{**/dist/**/index.module.js,./dist/index.css}"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,3 +12,4 @@ jobs:
       - uses: preactjs/compressed-size-action@v2
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          pattern: "**/dist/**/{index.module.js,index.css}"


omgovich:
Hey, @RyanChristian4427. Thanks for this contribution.

Look good but It seems to me that the action prints too many data. 
We don't need to compare all .js files — we could, for example, compare `**/*.module.js` only because they most probably will be imported to browser bundles. Then the script's output will be just 7 files.

In the action's docs, I found the way to customize the files list. Could you take a look?
https://github.com/preactjs/compressed-size-action#customizing-the-list-of-files

RyanChristian4427:
If we only look at `.module.js` then we'd miss potential regressions in say the UMD bundle. 

We should check them all as each are different targets and might use different polyfills.

Any values not changed will start minimized, for what that's worth.

omgovich:
I understand your position, but their sizes are connected anyway and `.module.js` output is the most important.
If you open `.umd.js` files you'll see that they're almost the same as `.module.js` ones but with additional export features in the beginning. So probably these extra lines won't be informative, but make PR pages longer =)

In my opinion, when someone creates a PR and a bot creates a huge 
We contribute for fun and because we love what we do, and I don't want to dramatically affect that by 40-rows reports from bots. I just feel that they will demotivate people.

I would keep things simpler and ask bots to print only important lines.

omgovich:
`*.js` size is not important since it's CJS for Node/SSR env mostly and for some legacy bundlers.
`*.esmodule.js` is times less important than `*.module.js` since it's is rarely used for production. It's the same as `*.module.js` so if we make it lighter, `*.esmodule.js` most probably became lighter too.
`*.umd.js` looks the same as `*.module.js` but with additional code for UMD export.

omgovich:
All I want to say that we could make things easier by just keeping an eye on `*.module.js` only.
This output format os the only that really important and it represents changes in modern and UMD bundles.

omgovich:
IMO watching for `dist/index.css` is way more important because developers have to load it no matter what color model they use.

RyanChristian4427:
I'm not sure you're going to scare anyone away with simple bundle numbers, but, your library. You likely will come across regressions though and I believe it would be better in that case to not to use this at all. It would only be introducing a false sense of security. There definitely are ways one particular bundle could grow much larger without `x.module.js` being effected at all.

That being said, looks like it will be easy enough to target just `x.module.js` builds along with the CSS.

.github/workflows/main.yml
```diff
@@ -0,0 +1,15 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: preactjs/compressed-size-action@v2
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          pattern: "**/dist/**/{index.module.js,index.css}"
```
omgovich, line 15: Since all index.css are equal files for now, could we check `./dist/index.css` only?
As far as I understand this should do the trick:
```
pattern: "{**/dist/**/index.module.js,./dist/index.css}"
```

omgovich: Then action output will be just 8 rows instead of 35+.

omgovich:
Thanks @RyanChristian4427 ❤️ I very appreciative the things you doing for the library.

PR #29, opened 2020-08-26
refactor(build-output): Moving all build output files into `x/dist/...`

RyanChristian4427:
Fixes a problem mentioned in #28, where an action would not work as it looks for changes in a `dist` directory.

Moving all of our build outputs inside of one is far more idiomatic and costs nothing. I don't see a reason to not do it. 

Previous structure:

```
rgb/
--package.json
--index.js
--index.umd.js
--...
```

New structure:

```
rgb/
--package.json
--dist/
----index.js
----index.umd.js
----...
```


omgovich:
I'm going to work on RGBA and HSLA color models that include a new UI element for change alpha channel, and probably these pickers will require their own stylesheet import. I mean:
```js
import RgbColorPicker from "react-colorful/rgba";
import "react-colorful/rgba/index.css";
```

I'm not sure that it will work this way, but if it will, I have to return to the current structure in order because `import "react-colorful/rgba/dist/index.css";` looks uglier.

RyanChristian4427:
One of your forks (on mobile, can't link) has added Alpha in some sort of way. Might want to check on them. 

As for CSS import path, you can just move the file after the fact.

omgovich:
Didn't know about this fork) Thanks.
I got your point about built file locations "standardization" in order to make it possible to measure their size by actions.
Ok, let's merge this. It doesn't affect anything now, so let's merge it then.

RyanChristian4427:
> I'm going to work on RGBA and HSLA color pickers that include a new UI element for change alpha channel, and probably these pickers will require their own stylesheet import. I mean:
> 
> ```js
> import RgbColorPicker from "react-colorful/rgba";
> import "react-colorful/rgba/index.css";
> ```
> 
> I'm not sure that it will work this way, but if it will, I have to return to the current structure because `import "react-colorful/rgba/dist/index.css";` looks uglier.

Sorry, little late here. 

Firstly, Microbundle will bundle all stylesheets together if I'm not mistaken. So unless you disable its handling of CSS (maybe just modules? not sure) you're never going to have a second stylesheet that an end user can import.

Secondly, why don't we just remove all stylesheets outside of the main export? Might need a few changes, but something like this:

```
dist/
--css/
----styles.css
--index.js
--...
```

Only `dist/` (the top-level one) will have any styles. So `rgb/`, `hsl/, etc. won't contain any CSS themselves.

Moving it into a `css/` directory I think is just a bit cleaner, even if that is another directory in the exported structure. Just my 2¢ though. 

Might need a little bit of Microbundle shenanigans but would be an improvement IMO.

omgovich:
@RyanChristian4427 Love the idea to have `css` folder. 
In order to avoid breaking changes, we should keep `dist/index.css`. We could just copy (not move) this file to achieve that.

RyanChristian4427:
Yeah, certainly. Just thinking of ways we could further reduce the complexity/duplication of the distribution. I'll have a play around at some point to see if there's an nice way to go about doing that.

PR #33, opened 2020-08-29
feat: Color object types as named exports

RyanChristian4427:
PR is in response to [this comment](https://github.com/omgovich/react-colorful/pull/23#issuecomment-683318037).

This adds HSL, HSV, and RGB as named exports for their respective color pickers, resulting in the following API:

```
import HslColorPicker, { HSL } from 'react-colorful/hsl';
```

This is a major improvement in accessibility over the previous solution:

```
import HslColorPicker from 'react-colorful/hsl';
import { HSL } from 'react-colorful/dist/types';
```

Only a single import statement required.

I also changed our types a bit to remove the indexing on the objects outside of the color object comparator scope. The color object comparator is a bit more verbose and more complicated with its casts but it's an improvement, especially with the types now being exported.

### DIFF
--- a/src/utils/compare.ts
+++ b/src/utils/compare.ts
@@ -2,15 +2,19 @@ import { hexToRgb } from "./convert";
 
 import { HSL, HSV, RGB } from "../types";
 
-interface ColorCompare {
-  [key: string]: number;
-}
-
 export const equalColorObjects = (first: HSL | HSV | RGB, second: HSL | HSV | RGB): boolean => {
   if (first === second) return true;
 
   for (const prop in first) {
-    if (((first as unknown) as ColorCompare)[prop] !== ((second as unknown) as ColorCompare)[prop])
+    // The following allows for a type-safe calling of this function (first & second have to be HSL, HSV, or RGB)
+    // with type-unsafe iterating over object keys. TS does not allow this without an index (`[key: string]: number`)
+    // on an object to define how iteration is normally done. To ensure extra keys are not allowed on our types,
+    // we must cast our object to unknown (as RGB demands `r` be a key, while `Record<string, x>` does not care if
+    // there is or not), and then as a type TS can iterate over.
+    if (
+      ((first as unknown) as Record<string, number>)[prop] !==
+      ((second as unknown) as Record<string, number>)[prop]
+    )
       return false;
   }
 
--- a/src/utils/compare.ts
+++ b/src/utils/compare.ts
@@ -2,11 +2,16 @@ import { hexToRgb } from "./convert";
 
 import { HSL, HSV, RGB } from "../types";
 
+interface ColorCompare {
+  [key: string]: number;
+}
+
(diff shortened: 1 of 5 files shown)


src/utils/compare.ts
```diff
@@ -2,11 +2,16 @@ import { hexToRgb } from "./convert";
...
+interface ColorCompare {
+  [key: string]: number;
+}
+
 export const equalColorObjects = (first: HSL | HSV | RGB, second: HSL | HSV | RGB): boolean => {
   if (first === second) return true;
 
   for (const prop in first) {
-    if (first[prop] !== second[prop]) return false;
+    if (((first as unknown) as ColorCompare)[prop] !== ((second as unknown) as ColorCompare)[prop])
```
omgovich, line 13: Uh, looks brutal, but as far as I understand there are no other options.

omgovich (approved):

src/utils/compare.ts
RyanChristian4427: Yeah, sorry.

I'll keep looking around, maybe there's a better solution I'm just not thinking of right now. 

Iterating over object keys is always going to be a bit awkward with TS though.

omgovich: We can merge this solution, but could I ask you to add a comment here describing what're we doing here and why?

RyanChristian4427: Absolutely. I can also remove the interface declaration and just in-line as `Record<string, number>` now they I think about it.

src/utils/compare.ts
```diff
@@ -6,7 +6,15 @@ export const equalColorObjects = (first: HSL | HSV | RGB, second: HSL | HSV | RG
...
   for (const prop in first) {
-    if (first[prop] !== second[prop]) return false;
+    // The following allows for a type-safe calling of this function (first & second have to be HSL, HSV, or RGB)
+    // with type-unsafe iterating over object keys. TS does not allow this without an index (`[key: string]: number`)
+    // on an object to define how iteration is normally done. To ensure extra keys are not allowed on our types,
+    // we must cast our object to unknown, and then as a type TS can iterate over.
+    if (
+      ((first as unknown) as Record<string, number>)[prop] !==
+      ((second as unknown) as Record<string, number>)[prop]
+    )
```
RyanChristian4427, line 16: @omgovich Is this along the lines of what you were thinking of?

omgovich: Yeah. Now everyone will be able to understand the magic that we're doing here =)

RyanChristian4427: Just added a bit more to why we need to first cast to `unknown` as that was lacking.

omgovich: ❤️

omgovich:
BTW: don't you think that we should mention in README that the color types are present for import?
I mean we could create something like "TypeScript support" section and add an expandable summary of available types. What do you think?

RyanChristian4427:
Sure, that would be useful.

PR #34, opened 2020-08-29
TS: Allow to pass another props to HexInput

omgovich:
@RyanChristian4427 I just noticed that it's not allowed to pass the default input tag props (such as `className`, `placeholder`, etc) to `HexInput` component.

### DIFF
--- a/src/components/HexInput.tsx
+++ b/src/components/HexInput.tsx
@@ -5,10 +5,9 @@ import { validHex } from "../utils/validate";
 // Escapes all non-hexadecimal characters including "#"
 const escape = (hex: string) => hex.replace(/([^0-9A-F]+)/gi, "");
 
-interface Props {
+interface Props extends HTMLInputElement {
   color: string;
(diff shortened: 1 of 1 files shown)


src/components/HexInput.tsx
```diff
@@ -8,6 +8,7 @@ const escape = (hex: string) => hex.replace(/([^0-9A-F]+)/gi, "");
 interface Props {
   color: string;
   onChange: (newColor: string) => void;
+  [key: string]: unknown;
```
RyanChristian4427, line 11: Sorry, could've sworn I fixed that as I remember doing this.

```suggestion
interface Props extends HTMLInputElement {
  color: string;
  onChange: (newColor: string) => void;
```

omgovich: That looks much better, thanks!

RyanChristian4427: Apologies, I know I tried this and used it, just don't know what happened to it.

PR #36, opened 2020-08-30
Rewrite the demo in TS

omgovich:
In order to keep the entire repo's codebase consistent.


demo/src/hooks/useFaviconColor.ts
```diff
@@ -75,13 +82,17 @@ const useFaviconColor = (color) => {
...
   };
 
-  useThrottledEffect(
-    () => {
-      if (shouldReplace()) replaceFavicon(color);
-    },
-    500,
-    [color]
-  );
+  // modified from `use-throttled-effect` since the lib doesn't support TS
```
RyanChristian4427, line 85: If you mean by this the library does not have typings, there are easy ways around this. Don't necessarily have to write up your own solution.

One way would be to use a declaration file.

`demo/src/declaration.d.ts`
```
declare module "use-throttled-effect";
```

That'll silence TS. Obviously won't have any typings, but you can then do what you want with it. Everything should be `any`. Alternatively you could write typings for this lib here, but that's likely overkill.

omgovich: Yeah, I know that I could define a module, but it seems like an extra work for a supersmall thing like this hook. It's seems easier to me to just copy and and add typings.

RyanChristian4427: Fair enough, just wanted to make sure you knew that they're was an easier solution (albeit with drawbacks)

omgovich:
I will close it for now since it will have too many conflicts with v3
I will come back to this project when we release a stable v3.

PR #38, opened 2020-08-31
docs(readme): Added Preact section

RyanChristian4427:
Adding documentation for usage with Preact.

Added a few links for common Preact starting points that will work right out of the box along with documentation on how to use when not using one of those starting points.

I removed "Preact Support" from the roadmap and labeled the Preact section as "Usage with Preact" because it's rather incorrect to say this library supports Preact in any way. It has been the Preact team's work through `preact/compat` that ensures most React libraries (this one included) will work flawlessly in a Preact application. Saying that we support it therefore feels a bit disingenuous as we really didn't do anything ourselves.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,12 @@ Take a look at [Supported Color Models](#supported-color-models) for more inform
 
 </details>
 
+## Preact Support
+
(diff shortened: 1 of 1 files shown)


omgovich:
Awesome! Thanks @RyanChristian4427!

PR #40, opened 2020-08-31
add couple improvements 🙌🏻

jeetiss:
Hey @omgovich,

I try do my best: 

- now only unique color fires `onChange` callback
- `onChange` callback is saved in ref, so `useCallback` is not required on parent level
- performance is improved

### DIFF
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -11,12 +11,25 @@ interface Props<T extends AnyColor> extends Partial<ColorPickerBaseProps<T>> {
   colorModel: ColorModel<T>;
 }
 
+function useEventCallback<T>(event?: (arg: T) => void) {
+  const eventRef = useRef(event);
+
+  useEffect(() => {
+    eventRef.current = event;
+  });
+
+  return useCallback((arg: T) => eventRef.current && eventRef.current(arg), []);
+}
+
 const ColorPicker = <T extends AnyColor>({
   className = "",
   colorModel,
   color = colorModel.defaultColor,
   onChange,
 }: Props<T>) => {
+  // Save onChange callback in the ref for avoiding "useCallback hell"
+  const onChangeRef = useEventCallback<T>(onChange);
+
   // No matter which color model is used (HEX, RGB or HSL),
   // all internal calculations are based on HSV model
   const [hsv, updateHsv] = useState<HSV>(() => colorModel.toHsv(color));
@@ -41,9 +54,9 @@ const ColorPicker = <T extends AnyColor>({
     const newColor = colorModel.fromHsv(hsv);
     if (!colorModel.equal(newColor, cache.current)) {
       cache.current = newColor;
-      if (onChange) onChange(newColor);
+      onChangeRef(newColor);
     }
-  }, [hsv, colorModel, onChange]);
+  }, [hsv, colorModel, onChangeRef]);
 
   // Merge the current HSV color object with updated params.
   // For example, when a child component sends `h` or `s` only
--- a/src/components/ColorPicker.tsx
+++ b/src/components/ColorPicker.tsx
@@ -5,7 +5,6 @@ import Saturation from "./Saturation";
 
 import styles from "../styles.css";
 import { ColorModel, HSV, ColorPickerBaseProps, AnyColor } from "../types";
-import { equalColorObjects } from "../utils/compare";
 import formatClassName from "../utils/formatClassName";
 
 interface Props<T extends AnyColor> extends Partial<ColorPickerBaseProps<T>> {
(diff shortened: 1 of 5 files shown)


omgovich:
Hi @jeetiss! Thanks for you contribution!
I'm on vacation, but I'll try to take a look ASAP. There are few thoughts that probably will make these improvements even better.

src/components/ColorPicker.tsx
```diff
@@ -5,46 +5,58 @@ import Saturation from "./Saturation";
...
 
+function useEventCallback<T>(event?: (arg: T) => void) {
+  const eventRef = useRef(event);
+
+  useEffect(() => {
+    eventRef.current = event;
+  });
+
+  return useCallback((arg: T) => eventRef.current && eventRef.current(arg), []);
+}
```
omgovich, line 22: Nice trick!

Shouldn't we do the same in `HexInput`?
I would move this function to `src/hooks/useEventCallback.js` in order to reuse it in both components.

src/components/ColorPicker.tsx
```diff
@@ -5,46 +5,58 @@ import Saturation from "./Saturation";
...
   // and save the new color to the ref to prevent unnecessary updates
   useEffect(() => {
-    if (!equalColorObjects(hsv, cache.current.hsv)) {
-      const newColor = colorModel.fromHsv(hsv);
-      cache.current = { hsv, color: newColor };
-      if (onChange) onChange(newColor);
+    const newColor = colorModel.fromHsv(hsv);
+    if (!colorModel.equal(newColor, cache.current)) {
+      cache.current = newColor;
+      onChangeRef(newColor);
```
omgovich, line 57: It seems to me that `onChangeRef` isn't the best name we could use here.
I mean, when most of the react-developers see a variable name that ends with `...Ref` they most probably expect that they have to add `.current`.

I would try something like `onChangeCallback(...)`. What do you think?

src/components/ColorPicker.tsx
```diff
@@ -5,46 +5,58 @@ import Saturation from "./Saturation";
...
+      cache.current = color;
       updateHsv(newHsv);
     }
   }, [color, colorModel]);
 
-  // If HSV is changed, convert it to the output format, send to the parent component
+  // Сonvert HSV to the output format, if it is changed send to the parent component
   // and save the new color to the ref to prevent unnecessary updates
   useEffect(() => {
-    if (!equalColorObjects(hsv, cache.current.hsv)) {
```
omgovich, line 42: I understand why you dropped `hsv` from `cache` and killed this condition.
This part of my code looks stupid but it makes sense =)

The problem is hidden inside the color conversion algorithms. They aren't 100% "accurate" for conditions we need here.
Mostly because of `Math.round` and `Math.floor` they have. Sometimes because of the conditions inside the code.

Just try to set `#c62182` as the default color in the demo and you'll see that in your branch `ColorPicker` triggers `onChange` right after the mounting and prints `#c72282` to the console. These colors are almost the same but they are not because during the `hex→rgb→hsv→rgb→hex` we run `Math.floor` and `Math.round` a lot of times and we are not able to get the source color. It's sad but that's how color conversion algorithms work.

```
const someHsv =  hexToHsv('#c62182')
hsvToHex('#c62182') // #c72282' - different color
```

So I store previous HSV in the ref to be able to compare them. Don't like this thing, but it works.

omgovich (changes requested):

src/components/ColorPicker.tsx
jeetiss: Nice name for var, renamed it )

tests/components.test.js
```diff
@@ -53,11 +53,31 @@ it("Accepts an additional `className`", () => {
...
+
+  rerender(<ColorPicker color="#c72282" onChange={handleChange} />);
+
+  expect(handleChange).not.toHaveReturned();
+});
+
+it("Doesn't trigger `onChange` when hex color is black", () => {
+  const handleChange = jest.fn();
+  const { container } = render(<ColorPicker color="#000" onChange={handleChange} />);
+  const hue = container.querySelector(".react-colorful__hue .interactive");
```
omgovich, line 73: Hmm. What's the goal of this test? Is there a case that other tests dont cover?

jeetiss: Yes ofc
When you change hue for `#000` color picker fires `onChange`, because `hsv(120 0 0)` -> `#000` and `hsv(0 0 0)` is black to. I added test fro this case.

omgovich: Ow, I see. Nice one!
Should we rename the test maybe?
Something like 
```
... if changing the hue doesn't change the hex color (grayscale colors)
```

omgovich: Or
```
Doesn't call `onChange` when user changes a hue of a grayscale color`
```

omgovich: Very nice test btw. Thank you!

jeetiss: I renamed test

src/components/ColorPicker.tsx
omgovich: @jeetiss could we reuse the hook inside of HexInput?

omgovich (approved):
Thanks a lot, @jeetiss!

PR #41, opened 2020-09-02
refactor: enhance type definition

byr-gdp:


src/components/HexInput.tsx
```diff
@@ -50,4 +50,4 @@ const HexInput = (props: Partial<Props>) => {
   return React.createElement("input", inputProps);
 };
 
-export default React.memo(HexInput);
+export default React.memo<Props>(HexInput);
```
molefrog, line 53: I guess, we should make the `Props` type partial here, since `onChage` and `color` are optional properties (see https://github.com/omgovich/react-colorful/blob/e2444497b4573fd6c023fcb6e177107b6991426f/src/components/HexInput.tsx#L13):
```suggestion
export default React.memo<Partial<Props>>(HexInput);
```

byr-gdp: Sounds reasonable, thanks~

RyanChristian4427: Not just `onChange` and `color`, but every key in `HTMLInputElement` as `Props` extends it.

RyanChristian4427:
@omgovich 

Certainly no rush, but can you see if you can restart the initial GitHub action? Looks to be stuck in the queue, unfortunately happens every now and then.  Not sure if it will post after a restart, but worth a shot.

omgovich:
Love the ideas in this PR. Thanks @byr-gdp.

I think we should make a separate PR later to improve event typing of the rest components (especially `Interaction`).

omgovich:
@RyanChristian4427 any suggestions regarding this PR? Looks good to me.

RyanChristian4427:
No, looks good to me.

Honestly not sure why strict TS lets us get away with un-typed events as we have. Automatically casts them to `any`. Definitely nice to have them strictly typed, lot of easy mistakes can be caught with this.

Great PR

PR #42, opened 2020-09-02
v3: Refactored package structure / module exports

RyanChristian4427:
As the title states, this is for a v3 down the road. Definitely needs a lot of feedback, thought, and approval before then though. 

I'll fill this out piece by piece, lot of documentation/justifications to write, but this is largely in response to https://github.com/omgovich/react-colorful/pull/23#issuecomment-685276064 along with a few other scattered discussions.

Basic idea is less tooling, better API.

We'll let modern bundlers do the treeshaking rather than trying to do it ourselves. Microbundle doesn't really support what we currently have which can be seen by the duplicated CSS and TS definition files, along with the build script. 

## End User Changes (and why)

I propose a few changes for the end user, namely:

- Different method of importing components
- (Maybe) Different route for importing CSS

1) Instead of doing as we were with default exports of sub-modules, we could be using combined named exports for the entire module. For example, 

    ```
    import ColorPicker from 'react-colorful';
    import HexInput from 'react-colorful/HexInput';
    ```

    becomes

    ```
    import { HexColorPicker, HexInput } from 'react-colorful';
    ```

    A bit less verbose, which could be nice, but overall similar. Breaking change, and will require users alter their imports.

    The reason for this is largely due to tooling, though the less verbose API certainly is convenient. With this setup our tooling responsibilities diminish greatly. A single entry point and a single bundle is what Microbundle does best. I'd also like to argue that named exports are far less ambiguous in terms of what they do vs. our current environment. 

    If I were to use `import ColorPicker from 'react-colorful';` then I'd have no idea what type `ColorPicker` was without looking at the documentation to see that `react-colorful`'s default export was the Hex color picker. Named exports cut right through this, while still offering options. `import { HexColorPicker } from 'react-colorful';` tells me exactly what kind of color picker I have, and I can use `{ HexColorPicker as ColorPicker }` if I so desired to keep the rest of my code the exact same.

2) In my mind, ideally, a CSS import route would look like: `import 'react-colorful/css/index.css';`. Not yet implemented in any way, but I do like the idea of moving the CSS outside of `dist/`. Makes for a cleaner path IMO. `import 'react-colorful/dist/index.css';` just seems like it isn't as nice.

## Internal Changes (and why)

- Single top-level entry point
- Moving of "package" files
- Removal of `build-package.js`

1) With a single top-level entry point our app will be easier to manage. All library exports (excluding css) will be in a single file, period. It reduces complexity as no longer will we have 7 separate files that decide what our output will look like. This is much simpler to deal with.

2) Our top-level for the library looks quite messy right now, solely due to the types targeting issues outlined here: https://github.com/developit/microbundle/issues/710#issue-685153433. However, with a top-level entry point like `index.ts`, we can easily target this for our `"types"` field resolving the issue. With that, I suggest we revert back to the original structure where the packages are nested away like the following:

    ```
    src/
    --index.ts
    --packages/
    ----hex.tsx
    ----hsv.tsx
    ----...
    ```

    This structure was much simpler and cleaner. 

3) The removal of `build-packages.js` comes as it is no longer needed. With a single entry point and named exports, Microbundle can handle the task easily on its own. This creates around a 4-5x increase in bundling speed. Obviously this isn't a magical speed increase, but due to the library changes. Still worth noting, faster build times == nicer experience, which is always a good thing.


package.json
```diff
@@ -2,17 +2,17 @@
...
   "esmodule": "dist/index.esmodule.js",
   "umd:main": "dist/index.umd.js",
-  "types": "dist/hex.d.ts",
+  "types": "dist/index.d.ts",
   "scripts": {
     "lint": "eslint src/**/*.{ts,tsx} demo/src/**/*.js",
     "size": "npm run build && size-limit",
     "test": "jest tests",
-    "build": "node ./build-packages.js",
+    "build": "microbundle build --entry src/index.ts --jsx React.createElement --css-modules true --sourcemap false --tsconfig tsconfig.build.json",
```
omgovich, line 15: as far as I know, we should add sideEffects key to package.json to mark the library as tree-shakable.
https://webpack.js.org/guides/tree-shaking/

omgovich (commented):
Nice!

package.json
RyanChristian4427: Definitely. We will have a side effect in the CSS but everything else should be side effect free.

demo/src/index.js
```diff
@@ -1,7 +1,6 @@
 import React, { useState, useEffect, useCallback } from "react";
 import ReactDOM from "react-dom";
-import ColorPicker from "../../src/hex";
-import HexInput from "../../src/HexInput";
+import { HexColorPicker, HexInput } from "../../src";
```
omgovich, line 3: **Discussion: Shouldn't we change the component names.**

These names seem inconsistent to me.
```
HexColorPicker
HexInput
```

At the same time, names like `RgbStringColorPicker` are too long IMO.

What do you think @RyanChristian4427. Is that important to have `Color` here or not?

Renaming everything this way feels right to me:
```
HexPicker
HexInput
RgbPicker
RgbStringPicker
...
```

react-color exports things like `CompactPicker`, `GithubPicker, `HuePicker`, etc. So I guess it's okay in terms of DX.

RyanChristian4427: Hm, I'd say I'm in favour of keeping `Color` in the names personally. While yes it adds some length, it also reduces ambiguity to 0. 

If I come across code that references a `HexColorPicker` I'd know and understand what that is instantly. `HexPicker` maybe not, at least until I look at the import to see where it's coming from.

RyanChristian4427: So I'd also be in favour of `HexColorInput`, but that's just me.

omgovich: I think you're right.
Let's rename `HexInput` to `HexColorInput` then?

src/index.ts
```diff
@@ -0,0 +1,17 @@
+import HexColorPicker from "./packages/hex";
+import HexInput from "./packages/HexInput";
+import HslColorPicker from "./packages/hsl";
+import HslStringColorPicker from "./packages/hslString";
+import HsvColorPicker from "./packages/hsv";
+import RgbColorPicker from "./packages/rgb";
+import RgbStringColorPicker from "./packages/rgbString";
```
omgovich, line 7: **Discussion: file names and structure:**

`packages` folder name seems outdated since we ship one tree-shakable package.
`hex.tsx` doesn't describe the file as a component code. 

My offer:
```
/src
  index.tsx
  
  /components
    HexColorPicker.tsx
    RgbColorPicker.tsx
    ...

    /common
      ColorPicker.tsx
      HexInput.tsx
      Hue.tsx
      ...
  
```

RyanChristian4427: Yeah, I agree, that's nicer.

RyanChristian4427:
I'll fix up some of these things when I can, but as I'm traveling at the moment, might be a while. I will come back and go through the comments w/ implementations when I get the chance though.

src/index.ts
```diff
@@ -0,0 +1,17 @@
+import HexColorPicker from "./packages/hex";
+import HexInput from "./packages/HexInput";
```
omgovich, line 2: **Discussion: imports and exports**

Shouldn't we migrate to named exports everywhere to make the codebase more consistent?

Bonus: we'll be able to replace these exports with:
```
export { HexColorPicker } from "./components/HexColorPicker";
import { HexColorInput } from "./components/HexColorInput";
```

RyanChristian4427: Internally: Could do. Not necessarily a problem to mix default with named though. Personally I usually do default exports with components as find that to be cleaner/simpler, but whatever you think.

For our exports, yeah, would make that much more concise.

src/index.ts
omgovich: Done

package.json
```diff
@@ -22,30 +22,6 @@
     {
       "path": "dist/index.module.js",
       "limit": "2 KB"
-    },
-    {
-      "path": "rgb/dist/index.module.js",
-      "limit": "2 KB"
```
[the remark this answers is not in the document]
omgovich: Done

package.json
omgovich: Done

demo/src/index.js
omgovich: Done

src/index.ts
omgovich: Done

omgovich:
@RyanChristian4427 Seems to me that it's done.
@molefrog Could you also take a look?

package.json
```diff
@@ -2,17 +2,18 @@
...
   "description": "A tiny color picker component for modern React apps",
-  "source": "src/hex.tsx",
+  "source": "src/index.ts",
   "main": "dist/index.js",
   "module": "dist/index.module.js",
   "esmodule": "dist/index.esmodule.js",
   "umd:main": "dist/index.umd.js",
-  "types": "dist/hex.d.ts",
+  "types": "dist/index.d.ts",
+  "sideEffects": false,
```
RyanChristian4427, line 11: I believe we need to label the CSS as a side-effect

```suggestion
  "sideEffects": [
    "*.css"
  ],
```

package.json
```diff
@@ -2,17 +2,18 @@
...
   "scripts": {
     "lint": "eslint src/**/*.{ts,tsx} demo/src/**/*.js",
     "size": "npm run build && size-limit",
     "test": "jest tests",
-    "build": "node ./build-packages.js",
+    "build": "microbundle build --entry src/index.ts --jsx React.createElement --name react-colorful --css-modules true --sourcemap false --tsconfig tsconfig.build.json",
```
RyanChristian4427, line 16: Do we really need to keep sourcemaps disabled? I was against it back when you did that, but figured I'd hold off as it was a fair bit of junk at the time.

But now there's just a few. Sourcemaps are quite useful when debugging.

package.json
```diff
@@ -62,14 +70,7 @@
     "ACKNOWLEDGMENTS",
     "LICENSE",
     "README.md",
-    "types",
```
RyanChristian4427, line 65: Good catch, I've missed that

src/components/HexColorInput.tsx
```diff
@@ -53,4 +52,4 @@ const HexInput = (props: Partial<Props>) => {
   return React.createElement("input", inputProps);
 };
 
-export default React.memo<Partial<Props>>(HexInput);
+export const HexColorInput = React.memo<Partial<Props>>(HexColorInputBase);
```
RyanChristian4427, line 55: Are we planning on using `HexColorInputBase` anywhere else? Could just wrap the main component in the `React.memo`. 

```suggestion
export const HexColorInput = React.memo<Partial<Props>>((...) => { ... });
```

README.md
```diff
@@ -124,29 +126,26 @@ The easiest way to tweak react-colorful is to create another stylesheet to overr
...
 <details>
-  <summary>How to use <code>HexInput</code></summary><br />
+  <summary>How to use <code>HexColorInput</code></summary><br />
 
-```diff
-import ColorPicker from "react-colorful";
-+import HexInput from "react-colorful/HexInput";
+```js
+import { HexColorPicker, HexColorInput } from "react-colorful";
 import "react-colorful/dist/index.css";
```
RyanChristian4427, line 133: I believe the CSS import is now incorrect because I started to mess with it before realizing we'd need something beyond Microbundle more than likely to do what I had imagined. Could just revert that though.

RyanChristian4427:
> 
> 
> @RyanChristian4427 Seems to me that it's done.
> @molefrog Could you also take a look?
> 
> **Pros:**
> 
>     * The library itself and especially the building process became way simpler and easier to maintain.
> 
> 
> **Cons:**
> 
>     * Import of `HexColorPicker` costs +100 bytes (probably because of named exports).
> 
>     * `HexColorInput` — 450 bytes → 1,25 KB (I don't know why yet)

Just out of curiosity, where are you getting the numbers from? I'll certainly have a quick look myself too when I get the chance.

package.json
RyanChristian4427: Actually no. Sorry. Was thinking of some other situation in which I had to list the CSS as a side effect but that shouldn't be applicable here.

README.md
```diff
@@ -124,29 +126,26 @@ The easiest way to tweak react-colorful is to create another stylesheet to overr
...
   );
 };
 ```
 
 [Live demo →](https://codesandbox.io/s/react-colorful-hex-input-demo-0k2fx)
 
-`HexInput` doesn't have any default styles, but accepts all properties that a regular `input` tag does (such as `className`, `placeholder` and `autoFocus`). That means you can place and modify this component as you like. Also, that allows you to combine the color picker and input in different ways.
-
-By the way, `HexInput` is also minimalist-friendly — only 400 bytes gzipped.
+`HexColorInput` doesn't have any default styles, but accepts all properties that a regular `input` tag does (such as `className`, `placeholder` and `autoFocus`). That means you can place and modify this component as you like. Also, that allows you to combine the color picker and input in different ways.
```
RyanChristian4427, line 148: I'm thinking the docs probably shouldn't be using contractions (I'm guilty of using them too) as it's less professional I guess? Totally a nitpick, but I'm going to run through quickly and fix that.

README.md
```diff
@@ -160,7 +159,7 @@ By the way, `HexInput` is also minimalist-friendly — only 400 bytes gzipped.
 While not only typing its own functions and variables, it can also help you type yours. Depending on the component you're using, you can also import the type that is associated with the component. For example, if you're using our HSL color picker component, you can also import the `HSL` type.
 
 ```ts
-import HslColorPicker, { HSL }  from "react-colorful/hsl";
+import { HslColorPicker, HSL } from "react-colorful";
```
RyanChristian4427, line 162: With named exports here do we want to find a new way to export the types? I was a fan of our old system as it pretty clearly distinguished the two but I think with named exports this will only get confusing. Now a type (`HSL`) is getting imported in the same way you'd import a component (`HslColorPicker`). 

Not sure I've seen another component library export the types, so I really don't know how we should handle this. New name for them that is less ambiguous, say `typeHsl`? Really not a fan of that name, but can't think of anything better right now.

The exported types are super convenient though, and I do think we should still export them.

README.md
```diff
@@ -28,7 +28,8 @@
 
 ## Features
 
-- **Small**: Just 1,5 KB (minified and gzipped). [Size Limit](https://github.com/ai/size-limit) controls the size.
+- **Small**: Just 1,6 KB (minified and gzipped). [Size Limit](https://github.com/ai/size-limit) controls the size.
+- **Tree-shakeable**: The only parts you need will be imported into your app's bundle.
```
RyanChristian4427, line 32:
```suggestion
- **Tree-shakeable**: Only the parts you use will be imported into your app's bundle.
```

I think this reads better

package.json
RyanChristian4427: With minified bundles I think sourcemaps are sort of required.

RyanChristian4427: Plus it's only weight in a dev environment, which will likely be 1000x the size as that's just the state of node. Losing a few kb from `node-modules` won't go noticed.

README.md
```diff
@@ -49,36 +50,37 @@ npm install react-colorful --save
...
 
 ```js
-import ColorPicker from "react-colorful";
+import { HexColorPicker } from "react-colorful";
 import "react-colorful/dist/index.css";
 
 const YourComponent = () => {
-  return <ColorPicker color={color} onChange={setColor} />;
+  return <HexColorPicker color={color} onChange={setColor} />;
```
RyanChristian4427, line 58: While we're talking about breaking changes, can I get your thoughts on naming that prop `color`?

It's happened a few times now, where I had to stop and think what a `color` property would be on a color picker. I'm wondering if something like `value` might be more idiomatic and easier to understand?

My problem with `color` is that, on a component, my first assumption would be that I'm setting some sort of styling for it. Maybe it's just me, but it takes a second to remember that `color` is the current value of the component. 

I personally think `value` would be more idiomatic, as then it matches more of the environment, for example, input fields. I feel that `value` and `onChange` fits better than `color` and `onChange`. While yes, it makes sense to set the color of a color picker, I think it's more accurate to say we're setting the current value of the color picker which just happens to be some representation of color. 

Just an idea I had.

package.json
omgovich: We dropped them off because the total size of NPM package was huge. Probably it's no longer a problem since we publish only one dist folder instead of 7.

omgovich:
> Just out of curiosity, where are you getting the numbers from? I'll certainly have a quick look myself too when I get the chance.

Run `npm run size`

README.md
omgovich: I had the same thought recently. The only thing that I'm afraid of is inconsistency with the color picker.

@molefrog We probably need 3rd mind here)

README.md
omgovich: Thanks!

README.md
omgovich: Do you want to change it for all components (the pickers and the input), right?

README.md
omgovich: I'm okay with the current state of color type exports.
We could rename them to make everything more consistent (internally and externally) by renaming them into `HsvColor, HslColor, RgbColor`.

It fits well with our `AnyColor` as well:

```
export type AnyColor = string | HslColor | HsvColor | RgbColor;
```

src/components/HexColorInput.tsx
omgovich: Understand what you mean but:
1) I don't like a mess like `React.memo<Partial<Props>>((props: Partial<Props>)`. It seems unreadable to me.
2) These things might throw ES lint warning about missing displayName.

package.json
omgovich: I got them back. The NPM package will be:
```
unpacked size: 156.9 kB
total files:   44
```

README.md
omgovich: Sorry, not sure that I got what you mean here.

README.md
RyanChristian4427: Yeah I like that. Fits better and I think it's also a more clear API.

README.md
RyanChristian4427: Not necessarily **want** to change it, but I had the thought that it could be more clear and idiomatic if we were to change it. 

But yes, for all the components if we were. 

What's the inconsistency?

omgovich: Forget it) I just thought you want to change `HexColorInput` only.

omgovich: I would keep `color` for now. It's okay I think and I don't want to make the PR bigger.

README.md
omgovich: Will do that then.

README.md
RyanChristian4427: I moved the CSS into `src/css/...`, so the output is now at `dist/css/index.css`. Will need fixing/reverting/something.

However, I still think it could be nice to, as an after-build process, move the CSS up a level, so `dist` and `css` will both be top-level for this package. This would allow a user to import CSS like `import "react-colorful/css/index.css"` which I think would be a bit nicer.

README.md
RyanChristian4427: Oh, gotcha. 

Yeah, fair, no problem.

README.md
omgovich: Done

README.md
RyanChristian4427: Does it really? Didn't know that. Would've assumed it kept the same directory structure. (Sorry, can't check right now as still traveling). In that case, less valuable. 

But sure, no reason to do anything further here on that.

omgovich: Actually we could even copy that to the root. Then it will be available by:
```
import "react-colorful/index.css"
or 
import "react-colorful/styles.css"
```

Anyway, I think this is topic for another discussion.

RyanChristian4427: Yeah so long as we're just copying it doesn't need to be a breaking change and can be done elsewhere at some other time.

omgovich (approved):
Seems to me that we're done here.
Everything looks super good. Like the result!
Thanks @RyanChristian4427!

omgovich:
Quick demo of v3: https://codesandbox.io/s/ts-test-pjyz1?file=/src/App.tsx

molefrog:
I'm little late to the party, but this looks good! 
```js
import { HexColorPicker } from "../../src/";
```

Does that mean we're moving away from the default color picker component? I wonder if we could make an extra export named just `ColorPicker` (an alias to `HexColorPicker`) for the sake of better UX?

omgovich:
Yeah, we moved away from default exports. It seems to me kinda wrong to mix named exports with the default one. I have heard that it could cause problems. So I would go with named exports only to avoid any issues in the future. I think the docs are clean enough.

molefrog:
@omgovich Having no default exports is fine and I do understand why it's better. But would that still be possible to do:
```js
import { ColorPicker } from "react-colorful"
```
(instead of importing a color picker specific to the color model, we could also provide a standard one, an alias to hex one, which will cover 90% of the use cases and potentially improve the experience of user who are just getting started with the library).

omgovich:
Probably it's a good idea, but I'm also afraid that it might confuse people who will read the docs.

package.json
omgovich: Hey @RyanChristian4427!
Just found out that you were right.
I added our picker to https://omatsuri.app and webpack removed all our styles from there.
See:
https://github.com/webpack/webpack/issues/8814#issuecomment-465223178

Released 3.0.2 patch with `"sideEffects": [ "*.css" ]` and now everything is fine.

omgovich:
@RyanChristian4427 Do you have a Twitter account?

RyanChristian4427:
@omgovich Actually, only just finally got around to making one. Nothing there yet, but I can be found at @_rschristian

PR #45, opened 2020-09-08
Refactor/misc

RyanChristian4427:
Just a couple of minor internal changes.

## Changes

- Added `del-cli` to our build command. Goes back to what we had in the build script.
- Correcting `ColorPickerBaseProps` type. It (incorrectly) did not extend `AnyColor`.
  - While it was very unlikely to ever cause an issue due to `ColorModel` constraining it, this can and should be tightened down further.
- Added `"noEmit": true` to our `tsconfig.json`. As we already use `microbundle` for all of our building, `tsc` doesn't really serve a purpose. However, if we enable `noEmit` then `tsc` can be used as a typechecker, which is quite useful. 
  - Use `yarn tsc` to check all types in the app without the production-level bundling. Much quicker when types are all that you want to check.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,7 +7,6 @@
     "jsxFactory": "React.createElement",
     "moduleResolution": "node",
     "noImplicitAny": true,
-    "outDir": "dist",
     "declaration": true,
     "strict": true,
     "noEmit": true
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,8 @@
     "noImplicitAny": true,
     "outDir": "dist",
     "declaration": true,
-    "strict": true
+    "strict": true,
+    "noEmit": true
   },
   "include": ["src", "test", "demo/src"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,8 +7,7 @@
     "jsxFactory": "React.createElement",
     "moduleResolution": "node",
     "declaration": true,
-    "strict": true,
-    "noEmit": true
+    "strict": true
   },
   "include": ["src", "test", "demo/src"]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,7 +6,6 @@
     "jsx": "react",
     "jsxFactory": "React.createElement",
     "moduleResolution": "node",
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -7,9 +7,9 @@
     "jsxFactory": "React.createElement",
     "moduleResolution": "node",
     "noImplicitAny": true,
-    "outDir": "dist",
     "declaration": true,
-    "strict": true
+    "strict": true,
+    "noEmit": true
```
omgovich, line 12: Seems kinda confusing to have this option here.
What if we define npm script like that:
```
"check-types": tsc --noEmit true"
```

It's also better because contributors will be able to find this command in package.json (no need to guess that `yarn tsc` exists)

tsconfig.json
```diff
@@ -7,9 +7,9 @@
     "jsxFactory": "React.createElement",
     "moduleResolution": "node",
     "noImplicitAny": true,
```
omgovich, line 9: BTW shouldn't we get rid of this `noImplicitAny` line?
AFAIK  `"strict": true` includes this statement:
image

RyanChristian4427: Yeah, that sounds right. 

It was just a copy/pasted tsconfig that I later added `strict` to.

tsconfig.json
RyanChristian4427: `tsc` is always going to be there with Typescript as a dependency though.

Separate script makes sense, but I do think we should keep `noEmit` as without it we'd essentially have two build tools at once. As `tsc` is commonly used in libs as a build tool, o think disabling that functionality of it might reduce chance of mistakes/confusion. 

I can see it either way though.

RyanChristian4427: Hm I don't think so? A bundler that would acknowledge TS's noEmit would be rather pointless... if noEmit is false, bundler acts normally. If noEmit true, do absolutely nothing? 

I can't see any reason why a bundler would acknowledge an option like that.

Fair enough though, I'll alter it.

omgovich:
Thanks @RyanChristian4427!

omgovich (approved):

PR #47, opened 2020-09-11
Feature: Alpha channel 

omgovich:
ezgif com-video-to-gif

This PR adds RGB**A**, HSL**A**, and HSV**A** color models support.

The internal color model now is HSV**A** (instead of HSV).

New public components:
```
RgbaColorPicker
RgbaStringColorPicker

HslaColorPicker
HslaStringColorPicker

HsvaColorPicker
```

New internal components:
```
Alpha
AlphaColorPicker
```

`AlphaColorPicker` helps me to keep old pickers (without alpha support) lighter. 
Only `AlphaColorPicker` imports `Alpha` component. Regular `ColorPicker` doesn't do that, so it's still light.

To avoid code duplication between these two components, I moved all logic to a hook (See `hooks/useColorManipulation.ts`).

### How to test

I added live examples of all picker components to the demo page.
Just run `npm run start-demo` and scroll down.

### Notes

Found out that `React.memo(Component)` makes a component "un-tree-shakeable".
That's the reason why `{ HexColorInput }` is so big.
So I got rid of some `React.memo`-s because they affect the budnle's size a lot. 

### WIP

There is an issue in terms of UI. This edge-state looks a bit weird:
image

Going to discuss that with our designer.

@molefrog @RyanChristian4427 Still WIP, but feedback welcome =)


RyanChristian4427:
Any reason for not adding Alpha with Hex?

omgovich:
> Any reason for not adding Alpha with Hex?

`{ HexColorPicker }` import size:

**master**:
1,6 KB

**this branch**:
1.71 KB

If I include `Alpha` into `ColorPicker` under condition like `{colorModel.hasAlpha && <Alpha ... />}`:
1.93 KB

Bundlers can't tree-shake it.

RyanChristian4427:
Sure, but what if we add a separate Hex with Alpha component? Seems a bit odd to add Alpha for all but 1.

omgovich:
We have 11 color pickers now:
```

// These are based on `ColorPicker.tsx` because their input/output doesn't have alpha-channel.
export { HexColorPicker } from "./components/HexColorPicker";
export { RgbColorPicker } from "./components/RgbColorPicker";
export { RgbStringColorPicker } from "./components/RgbStringColorPicker";
export { HslColorPicker } from "./components/HslColorPicker";
export { HslStringColorPicker } from "./components/HslStringColorPicker";
export { HsvColorPicker } from "./components/HsvColorPicker";

// These are based on `AlphaColorPicker.tsx` which includes `Alpha.tsx` component
export { RgbaColorPicker } from "./components/RgbaColorPicker";
export { RgbaStringColorPicker } from "./components/RgbaStringColorPicker";
export { HslaColorPicker } from "./components/HslaColorPicker";
export { HslaStringColorPicker } from "./components/HslaStringColorPicker";
export { HsvaColorPicker } from "./components/HsvaColorPicker";
```

HEX format doesn't need the alpha channel support.
There is HEX8 format, but I'm not going to support it.

omgovich:
Do you mean that we should support HEX8 format?

RyanChristian4427:
The overall amount of components we have doesn't really matter. Even if we had 30 the library is just as tiny at point of use. 

But yes, I was talking about 4/8 digit Hex, which is Hex with Alpha. If we're adding Alpha support to all other color pickers we should probably add it to Hex too while we're at it. It would be odd not to.

RyanChristian4427:
> ```js
> // These are based on `ColorPicker.tsx` because their input/output doesn't have alpha-channel (like `rgb(0, 0, 0)`)
> // They don't need `Alpha` component (extra 200+ bytes gz)
> export { HexColorPicker } from "./components/HexColorPicker";
> export { RgbColorPicker } from "./components/RgbColorPicker";
> export { RgbStringColorPicker } from "./components/RgbStringColorPicker";
> export { HslColorPicker } from "./components/HslColorPicker";
> export { HslStringColorPicker } from "./components/HslStringColorPicker";
> export { HsvColorPicker } from "./components/HsvColorPicker";
> 
> // These are based on `AlphaColorPicker.tsx` which includes `Alpha.tsx` component for output like `rgba(0, 0, 0, 1)`
> export { RgbaColorPicker } from "./components/RgbaColorPicker";
> export { RgbaStringColorPicker } from "./components/RgbaStringColorPicker";
> export { HslaColorPicker } from "./components/HslaColorPicker";
> export { HslaStringColorPicker } from "./components/HslaStringColorPicker";
> export { HsvaColorPicker } from "./components/HsvaColorPicker";
> ```

Looking at this reminds me, any reason for there to be no `HsvStringColorPicker`?

omgovich:
I don't know why. I just thought that it's not super important for now.
I also wanted to add CMYK format support, but postpone it while we have more important stuff to do.

Let's skip `HEX8` and `HsvString` in this PR.
The first one is tricky to add since it might break our primary HEX color model. It's definitely a separate task.
I will `HsvString` later in a separate PR. Just don't want to make this PR even bigger then it is already =)

RyanChristian4427:
Fair. 

The HsvString is totally for another PR, just occurred to me while looking over your comment. It would have 0 connection to this PR. Hex would, but yeah, might be bit complicated.

README.md
```diff
@@ -61,21 +61,26 @@ const YourComponent = () => {
...
+| `{ RgbColorPicker }`        | `{ r: 255, g: 255, b: 255 }`       | ~1,6 KB        |
+| `{ RgbaColorPicker }`       | `{ r: 255, g: 255, b: 255, a: 1 }` | ~1,7 KB        |
+| `{ RgbStringColorPicker }`  | `"rgb(255, 255, 255)"`             | ~1,7 KB        |
+| `{ RgbaStringColorPicker }` | `"rgba(255, 255, 255, 1)"`         | ~1,9 KB        |
+| `{ HslColorPicker }`        | `{ h: 0, s: 0, l: 100 }`           | ~1,4 KB        |
+| `{ HslaColorPicker }`       | `{ h: 0, s: 0, l: 100, a: 1 }`     | ~1,6 KB        |
+| `{ HslStringColorPicker }`  | `"hsl(0, 0%, 100%)"`               | ~1,5 KB        |
+| `{ HslaStringColorPicker }` | `"hsla(0, 0%, 100%, 1)"`           | ~1,7 KB        |
+| `{ HsvColorPicker }`        | `{ h: 0, s: 0, v: 100 }`           | ~1,4 KB        |
+| `{ HsvaColorPicker }`       | `{ h: 0, s: 0, v: 100, a: 1 }`     | ~1,5 KB        |
```
RyanChristian4427, line 83: I think we should just call the lib sub-2kb and call it a day. Adding sizes for each component, especially now that there are 10, just seems excessive. I don't think any user would be okay with say a 1.4kb component but totally against 2kb. Therefor I doubt removing the size will turn anyone away.

omgovich: You're right. It's kinda PITA =)

src/components/HslaStringColorPicker.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+import { AlphaColorPicker } from "./common/AlphaColorPicker";
+import { ColorModel, ColorPickerBaseProps } from "../types";
+import { equalColorString } from "../utils/compare";
+import { hslaStringToHsva, hsvaToHslaString } from "../utils/convert";
+
+const colorModel: ColorModel<string> = {
+  defaultColor: "hsl(0, 0%, 0%, 1)",
```
web-padawan, line 9: Looks like `hsvaToHslaString` uses `hsla()` format, but the default value here is in `hsl` format.

src/components/RgbaStringColorPicker.tsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+
+import { AlphaColorPicker } from "./common/AlphaColorPicker";
+import { ColorModel, ColorPickerBaseProps } from "../types";
+import { equalColorString } from "../utils/compare";
+import { rgbaStringToHsva, hsvaToRgbaString } from "../utils/convert";
+
+const colorModel: ColorModel<string> = {
+  defaultColor: "rgb(0, 0, 0, 1)",
```
web-padawan, line 9: Same as above: `hsvaToRgbaString ` uses `rgba()` format, but the default value here is in `rgb` format.

web-padawan (commented):
Two minor comments regarding formats used in default colors. 

In case if you don't support IE11, it could be fine to always use `rgb` and `hsl`, respectively.
According to caniuse, IE11 is the only browser that [does not support](https://caniuse.com/mdn-css_types_color_rgb_function_accepts_alpha) alpha in `rgb` / `hsl`.

src/components/RgbaStringColorPicker.tsx
omgovich: Good catch! Thx!

PR #50, opened 2020-09-14
Improve `Interactive` internal typing

omgovich:
@RyanChristian4427 Could you take a look?


RyanChristian4427:
LGTM 👍

Sorry again about missing a few of these types. I usually rely on Webstorm to tell me when types cannot be inferred. Unfortunately React casts these callbacks to `any` so no warnings ever popped up.

PR #53, opened 2020-09-14
fix: Resolving demo errors on Firefox

### LINKED ISSUE #52: Demo throws errors and is non-functional
@omgovich 

Seems like your latest deploy does not work on Firefox, tested both Windows & Linux (Manjaro, Arch). Looks like we need an additional check around [`TouchEvent`](https://stackoverflow.com/questions/27313488/touchevent-not-working-in-firefox-and-other-website-browser)

RyanChristian4427:
Fixes #52 

Demo on FireFox (Windows and Linux) throws errors that `TouchEvent` is not defined when attempting to move the picker. The fix is just a check to see if it's defined.


omgovich:
Thanks a lot @RyanChristian4427!

PR #56, opened 2020-09-16
Fix: Fast Tap and Release in iOS Safari

### LINKED ISSUE #55: Fast Tap and Release on Alpha or Hue Slider Causes Incorrect Selection
Noticed a small issue on iOS (13.7, tested on iPad and iPhone) that feels like a debounced input issue. To reproduce, start a tap and drag **away from the scrubber** and then release quickly. You’ll see the color revert to the initial tap, not the final drag. [Here’s a video of how it looks](https://i.imgur.com/gNlVJx7.gifv).

Also, I’m a big fan of this library. Surprising how few options there are out there and none of them at a reasonable size.

omgovich:
Fixes #55

The problem happens because sometimes iOS Safari calls `mousedown` and `mouseup` right after short touch interactions.
It seems to me that the browser thinks that it was a click (a combination of `mousedown` and `mouseup`) so it handles these events as well.

### DIFF
--- a/src/components/common/Interactive.tsx
+++ b/src/components/common/Interactive.tsx
@@ -6,6 +6,9 @@ import styles from "../../css/styles.css";
 // Use ternary operator instead of `Math.min(Math.max(0, number), 1)` to save few bytes
 const limit = (number: number) => (number > 1 ? 1 : number < 0 ? 0 : number);
 
+// Check if an event was triggered by touch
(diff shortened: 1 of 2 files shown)


molefrog (approved):
Smart! Any possibility to write a regression test to make sure it never breaks again?
Should be possible with `FakeMouseEvent`, but not 100% sure.

omgovich:
Good idea @molefrog. Added a simple test for that.

PR #58, opened 2020-09-18
Add a README section with project ports

molefrog:
Not sure about the header though. Alternatives:
  - "Other Frameworks"
  - "Ports"

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ const YourComponent = () => {
 
 <details>
   <summary>How you can get the most from our TypeScript support</summary><br />
-  
+
 While not only typing its own functions and variables, it can also help you type yours. Depending on the component you are using, you can also import the type that is associated with the component. For example, if you are using our HSL color picker component, you can also import the `HSL` type.
 
 ```ts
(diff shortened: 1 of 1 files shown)


omgovich:
What if we go with something like `Not using react?` section and write there about Preact and vanilla-colorful?

rschristian:
I think what @molefrog wrote makes sense. Preact should be an environment a React component can be dropped into without issue most of the time. No need to separate out Preact as `react-colorful` should operate in the very same fashion as it does in React. 

This, of course, is ignoring the current typing issue when using `@types/react` + TS Preact, though TS is always going to have some difficulties in that realm until they allow for aliasing.

omgovich:
Are we really expect to have any other port? I think we could just write that we have a framework-agnostic port and that's it =) I'm just afraid that the README is too long already. So I always try to keep things short or collapse them into `details` tag.

rschristian:
I don't necessarily disagree, I just think 'Not using React/Preact?' is fine as-is as there are no more instructions we can provide at least on Preact's behalf. Either it works as it does with React or it's a bug for one of the libs. We don't really have more instructions to give.

While I've written about the aliasing for Preact the reality of it is that if anyone is using Preact without the Preact CLI or another solution with already set up aliases then they know what they're doing as we don't have to remind them. 

As for the ReadMe, I don't think anyone is particularly bothered about ReadMe length. It's better to have more info in the ReadMe than to force users to read through the source / raise an issue in order to get the information they want.

Personally, I think the collapsed tags are a mistake as they only increase the difficulty of reading the document. You can't use `ctrl+f` to search if they're not expanded, they're not widely or often used, and browsers/GitHub don't exactly make it clear that those are elements that can be clicked on. At least with FireFox on Windows and Linux there's little indication at all that those are elements to be expanded.

The `ctrl+f` is the big one though. Users need to manually expand elements before they can effectively search the page. I'd be all for less nesting.

molefrog:
@omgovich Can we maybe merge the current version if the spelling/grammar looks ok?

PR #59, opened 2020-09-20
Refactor/built types

rschristian:
### Changes

This PR removes 12 `.d.ts` files that are internal types and not exposed to the end users. This brings the total number of files in the package from 41 down to 29. 

I moved this to a `.npmignore` file as I couldn't get the `files` field to cooperate with wildcards. Not entirely sure what was going on there, as the docs say it should be supported, but I figure this is a fine solution. Having a dedicated file for what is included and excluded in the bundle could be convenient.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,10 @@
+# blacklist all
+**
+
+# include whitelist
+!dist/*
+!dist/components/*
+!LICENSE
+!ACKNOWLEDGMENTS
+!README.md
+!package.json
--- a/.gitignore
+++ b/.gitignore
@@ -14,15 +14,6 @@ node_modules/
 .cache
 dist/
 
-# packages
-/hex
-/hsl
-/hslString
-/hsv
-/rgb
-/rgbString
-/HexInput
-
 # OSX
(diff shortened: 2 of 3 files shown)


.npmignore
```diff
@@ -0,0 +1,10 @@
+# blacklist all
+**
+
+# include whitelist
+!dist/*
+!dist/components/*
+!LICENSE
+!ACKNOWLEDGMENTS
+!README.md
+!package.json
```
rschristian, line 10: While `package.json` is one of the files that are always included, as are `README.md` and `LICENSE`, I figure having them typed out here only helps with clarity.

omgovich: Hi @rschristian. Thanks.
I on a vacation: will test as soon as I have time.

Have you run `npm publish --dry-run`?

rschristian: No problem,enjoy your vacation. Nothing pressing about this. 

Yes, I have tested with `publish --dry-run`. That's where I got the file count from.

omgovich:
Looks good! Thanks @rschristian!

PR #63, opened 2020-09-23
Accessibility, keyboard support

### LINKED ISSUE #57: Implement accessibility based on ARIA practices
Hello 👋   
I've been thinking about making the color picker accessible, and I would like to discuss the related changes.

It looks like for `Hue` and `Alpha` sliders we can follow [WAI-ARIA authoring practices](https://www.w3.org/TR/wai-aria-practices/#slider). See also the [example](https://www.w3.org/TR/wai-aria-practices/examples/slider/slider-1.html).
I'm not an expert, but presumably the same approach should work for `Saturation` as well.

There are following open questions that come to my mind:

1. Implementing this would require to figure out what the proper `step` value per arrow key press should be.
Should we make this step configurable? Maybe not as pubic property, but a protected one.

2. Note that `Page Up` and `Page Down` allow to change value by multiple steps (10 steps in the example).
Should this be configurable as well?

3. Proper support for screen reader announcements would require to provide a way for setting `aria-label` on the pointer elements. It could be done either with individual props or by passing an object with the following structure:

```js
i18n: {
  alpha: 'Alpha',
  hue: 'Hue',
  saturation: 'Saturation'
}
```

I'm going to prototype some of these changes in `vanilla-colorful` so I will create an issue there as well.

omgovich:
Closes #57

### DIFF
--- a/src/components/common/Alpha.tsx
+++ b/src/components/common/Alpha.tsx
@@ -16,12 +16,12 @@ interface Props {
 }
 
 export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
-  // Alpha always fit into [0, 1] range
   const handleMove = (interaction: Interaction) => {
     onChange({ a: interaction.left });
   };
 
   const handleKey = (offset: Interaction) => {
+    // Alpha always fit into [0, 1] range
     onChange({ a: clamp(hsva.a + offset.left) });
   };
 
--- a/src/components/common/Alpha.tsx
+++ b/src/components/common/Alpha.tsx
@@ -27,6 +27,8 @@ export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
     });
   };
 
+  // We use `Object.assign` instead of the spread operator
+  // to prevent adding the polyfill (about 150 bytes gzipped)
   const colorFrom = hsvaToHslaString(Object.assign({}, hsva, { a: 0 }));
   const colorTo = hsvaToHslaString(Object.assign({}, hsva, { a: 1 }));
 
--- a/src/components/common/Alpha.tsx
+++ b/src/components/common/Alpha.tsx
@@ -22,9 +22,7 @@ export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
   };
 
   const handleKey = (offset: Interaction) => {
-    onChange({
-      a: limit(hsva.a + offset.left),
-    });
+    onChange({ a: limit(hsva.a + offset.left) });
   };
 
   // We use `Object.assign` instead of the spread operator
--- a/src/components/common/Alpha.tsx
+++ b/src/components/common/Alpha.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from "react";
+import React from "react";
 
 import { Interactive, Interaction } from "./Interactive";
 
@@ -16,13 +16,10 @@ interface Props {
 }
 
 export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
-  const handleMove = useCallback(
-    (interaction: Interaction) => {
-      // Alpha always fit into [0, 1] range
-      onChange({ a: interaction.left });
-    },
-    [onChange]
-  );
+  // Alpha always fit into [0, 1] range
+  const handleMove = (interaction: Interaction) => {
+    onChange({ a: interaction.left });
+  };
 
   const handleKey = (offset: Interaction) => {
     onChange({
--- a/src/components/common/Alpha.tsx
+++ b/src/components/common/Alpha.tsx
@@ -27,8 +27,8 @@ export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
     });
   };
 
-  const colorFrom = hsvaToHslaString({ ...hsva, a: 0 });
-  const colorTo = hsvaToHslaString({ ...hsva, a: 1 });
+  const colorFrom = hsvaToHslaString(Object.assign({}, hsva, { a: 0 }));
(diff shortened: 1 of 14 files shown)


omgovich:
Had to apply a few tricks to keep all pickers inside 2 KB size limit.

omgovich:
@molefrog @rschristian Hey guys! Could I ask you to take a look?

omgovich:
HexColorPicker size:
```
1.51 → 1.86 KB
```

I added a lot of new code, so I think it's a pretty good result.

src/utils/limit.ts
```diff
@@ -0,0 +1,5 @@
+// Limit number within bounds.
+// Used ternary operator instead of `Math.min(Math.max(min, number), max)` to save few bytes
+export const limit = (number: number, min = 0, max = 1): number => {
```
molefrog, line 3: Imo, there is a better name for this method: `clamp`, used widely in animations (react-spring) and [the latest CSS proposal](https://developer.mozilla.org/en-US/docs/Web/CSS/clamp).

omgovich: Agree

omgovich: Done in 99be541

omgovich:
Added few specs to test keyboard support.

src/utils/clamp.ts
```diff
@@ -0,0 +1,5 @@
+// Clamps a value between an upper and lower bound.
+// Used ternary operator instead of `Math.min(Math.max(min, number), max)` to save few bytes
+export const clamp = (number: number, min = 0, max = 1): number => {
+  return number > max ? max : number < min ? min : number;
```
rschristian, line 4:
```suggestion
  return Math.max(min, Math.min(number, max));
```

FWIW, I think this is like ~10-20b smaller. Counting pennies, but personally I think it's more readable too.

src/components/common/Alpha.tsx
```diff
@@ -13,16 +16,19 @@ interface Props {
...
-    [onChange]
-  );
+  // Alpha always fit into [0, 1] range
+  const handleMove = (interaction: Interaction) => {
+    onChange({ a: interaction.left });
+  };
+
+  const handleKey = (offset: Interaction) => {
+    onChange({ a: clamp(hsva.a + offset.left) });
+  };
```
rschristian, line 26: Should the comment (`// Alpha always...`) be moved down to be above the usage of `clamp()`?

src/utils/convert.ts
```diff
@@ -32,7 +32,7 @@ export const hslaStringToHsva = (hslString: string): HsvaColor => {
     h: Number(match[1]),
     s: Number(match[2]),
     l: Number(match[3]),
-    a: match[4] !== undefined ? Number(match[4]) : 1,
+    a: match[4] == null ? 1 : Number(match[4]),
```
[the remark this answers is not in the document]
omgovich: It's just shorter version of `x === undefined`. Saves few bites.

See:
image

src/utils/clamp.ts
omgovich: In this project, the output size is more important than the source code style. 
We use tricks and they make code worse-looking but we cover then with comments.

src/utils/convert.ts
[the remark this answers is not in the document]
omgovich: But you're right it's too confusing. I returned the previous version (with `=== undefined`).

src/utils/clamp.ts
rschristian: Yes, what I'm saying is that the suggestion I provided is 10-20b lighter. 

[Existing outputs to 207 bytes](https://babeljs.io/repl#?browsers=IE%2011&build=&builtIns=usage&spec=false&loose=true&code_lz=MYewdgzgLgBMA2BDAtgBxgXhgCjAV2QCMBTAJwBoZkBLMTGABkuUQA96BGASkwD4YA3gCgYMUsSh5SdALKIoACwB0LVthphKcxStq4CJClTZcuAbiEBfIUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=true&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cenv&prettier=false&targets=&version=7.11.6&externalPlugins=)

[My suggestion outputs to 195](https://babeljs.io/repl#?browsers=IE%2011&build=&builtIns=usage&spec=false&loose=true&code_lz=MYewdgzgLgBMA2BDAtgBxgXhgCjAV2QCMBTAJwBoZkBLMTGABkuUQA96BGASkwD4YA3gCgYMUsSh5SdALKIoACwB0LVthphKcxStq4CJClTZcuAbiEBfIUA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=true&timeTravel=false&sourceType=module&lineWrap=false&presets=env%2Cenv&prettier=false&targets=&version=7.11.6&externalPlugins=)

rschristian: So you shave off 12b using `Math.max()` and the like

omgovich: Got you. But you should compare minified code. Solution based on ternary operators is almost 2 times shorter.
```js
Math.max(i,Math.min(n,a));
n>a?a:n<i?i:n;
```

rschristian: Ah, true enough.

omgovich: Described this thing in 6e810e6

rschristian: Cheers. Sorry about that.

molefrog (approved):
Great work!

omgovich:
Thanks, @molefrog @rschristian!

PR #65, opened 2020-09-25
Fix IE11 support

### LINKED ISSUE #37: Internet explorer support
Currently examples are not working on internet explorer 11.
Are there any plans for supporting it?

omgovich:
Closes #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ const YourComponent = () => {
 
 </details>
 
-## TypeScript Support
+## TypeScript support
 
 **react-colorful** supports TypeScript and ships with types in the library itself; no need for any other install.
 
@@ -226,6 +226,11 @@ If your port is not in the list, reach us out via [GitHub issues](https://github
     <img src="demo/src/assets/omatsuri-app.png" width="1223" alt="omatsuri.app" />
   </a>
 </details>
+
+## Browser support
+
+It's easier to say that **react-colorful** works everywhere. But if you need more detailed information, we have tested our picker in the browsers that were released in 2014–2015: Chrome 40, Firefox 32, Opera 25, Safari 9 and Edge 12.
+The picker works even in IE11 if an app has an `Object.assign` polyfill installed.
 
 ## Roadmap
 
--- a/README.md
+++ b/README.md
@@ -229,7 +229,10 @@ If your port is not in the list, reach us out via [GitHub issues](https://github
 
 ## Browser support
 
-**react-colorful** aims to support all browsers on the market and we test against legacy versions of many of the popular browsers. **react-colorful** will work out-of-the-box with all modern browsers and only requires that a `Object.assign` polyfill be provided for full IE11 support.
+It would be an easier task to list all of the browsers and versions that **react-colorful** does not support! We regularly test against browser versions going all the way back to 2013 and this includes IE11.
+
+**react-colorful** works out-of-the-box for most browsers, regardless of version, and only requires an `Object.assign` polyfill be provided for full IE11 support.
+
 ## Roadmap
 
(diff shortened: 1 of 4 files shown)


omgovich:
@rschristian Could I ask to proofread the new section? Probably you will find better words to describe that.

README.md
```diff
@@ -227,6 +227,11 @@ If your port is not in the list, reach us out via [GitHub issues](https://github
   </a>
 </details>
 
+## Browser support
+
+It's easier to say that **react-colorful** works everywhere. But if you need more detailed information, we have tested our picker in the browsers that were released in 2014–2015: Chrome 40, Firefox 32, Opera 25, Safari 9 and Edge 12.
+The picker works even in IE11 if an app has an `Object.assign` polyfill installed.
+
```
rschristian, line 234: I might reword it like this:

```suggestion
**react-colorful** supports all modern browsers (Chrome, Firefox, Opera, Safari, Edge) and IE11. It will work out of the box with modern browsers and only requires a `Object.assign` polyfill be provided for IE11 support. 
```

I think that saying it "works everywhere" is a bit of a stretch. Unless you're planning on supporting IE6 the scope should probably be limited down to the modern-ish options.

Furthermore I don't think there's a real need to list out what versions are tested against. I think it's incredibly unlikely for a user to be using Firefox or Chrome but using version 20. I think that saying we support the modern browsers is enough detail in that regard.

omgovich: I want to create a feeling that the picker works not only in modern browsers. That's why I avoided the "modern" word.
I mean when someone opens our README, he will see that our size is less than 2 KB and decided that we support only the latest versions, but that's not true.

omgovich: I would mention somehow that the picker works even in browsers that were released in 2015. I think that makes developers more confident that they can use our components in their production projects.

omgovich: Most of the **tiny** frontend projects work only in the modern browsers and don't even support, for example, Edge 15. It would be cool to describe that we support a lot of browsers.

rschristian: I guess that comes down to reader. I see "modern browser" to mean Chrome, Firefox, etc. regardless of version. Even Chrome 40 is "modern". But I see what you mean, there's no reason to leave that up to interpretation. How's this? 

```suggestion
**react-colorful** aims to support all browsers on the market and we test against legacy versions of many of the popular browsers. **react-colorful** will work out-of-the-box with all modern browsers and only requires that a `Object.assign` polyfill be provided for full IE11 support.
```

rschristian: > 
> 
> I would mention somehow that the picker works even in browsers that were released in 2015. I think that makes developers more confident that they can use our components in their production projects.

Personally I think listing years support just gets a bit too "down in the details" for what should likely be a high-level overview. Saying exactly what versions are supported, especially with some of these that are on v80+, feels a bit too specific.

omgovich: Looks good! Thanks ❤️

rschristian: Or maybe something like this:

```suggestion
It'd be an easier task to list all of the browsers and versions that **react-colorful** does not support! We regularly test against browser versions going all the way back to 2013, and this includes IE11. 

**react-colorful** works out-of-the-box for most browsers, regardless of version, and only requires a `Object.assign` polyfill be provided for full IE11 support. 
```

I think 2013 is a better number due to the IE11 launch date. I doubt anyone's using Firefox or Chrome versions that old.

rschristian: Bit more of a sales pitch than the previous one

omgovich: Love the last text a lot!

omgovich: Exactly what I wanted)

omgovich:
Thanks, @rschristian! I'm so lucky that you're on the team.

PR #66, opened 2020-09-26
docs(readme): Expounding upon Preact + TS set up

### LINKED ISSUE #39: Types: React-Colorful will not work with the TypeScript Preact-CLI template
If a user were to create a new TypeScript Preact project using the Preact CLI (`npx preact-cli create typescript preact-test`) and use this library following the instructions, they will run into the following error:

```
TS2786: 'ColorPicker' cannot be used as a JSX component.
  Its return type 'JSX.Element' is not a valid JSX element.
    Type 'Element' is missing the following properties from type 'Element': nodeName, attributes, children
```

Not entirely sure where it's coming from. I have a personal project that I was testing with while doing the TS rewrite and it still runs without any issues. Adding `react-colorful` to the base Preact template seems to be an issue though. Odd. 

Likely some issue with the typing differences between React & Preact but it might be possible to resolve by altering our typings slightly.

rschristian:
Would close #39 

This PR adds a subsection to the Preact section of the ReadMe that describes a potential problem that can arise with a Preact TS application and this library (or any React component library that uses `@types/react` for that matter). 

As the [Preact TypeScript template](https://github.com/preactjs-templates/typescript) will throw this error right out-of-the-box I feel that it is important to note, even if not an issue that this library can solve.

As a heavy user of Preact, I love when a library author acknowledges Preact users and notes any potential changes that will need to be made. I figure we can do the same here especially as this is quite an easy fix.

I'm not entirely sure that I love the wording right now, might come back later and make some adjustments or rewrite completely. 

## Changes
* Add documentation for Preact + TS

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,31 @@ Take a look at [Supported Color Models](#supported-color-models) for more inform
 
 If you are using another solution, please refer to the [Aliasing React to Preact](https://preactjs.com/guide/v10/getting-started#aliasing-react-to-preact) section of the Preact documentation.
 
+<details>
+  <summary>Preact + Typescript</summary><br />
+
+**react-colorful**, like all other React + TS projects, can potentially cause issues in a Preact + TS application if you have the `@types/react` package installed, either as a direct dependency or a dependency of a dependency. For example, the Preact TS template comes with `@types/enzyme` which has `@types/react` as a dependency.
+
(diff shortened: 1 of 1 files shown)


rschristian:
@betweenvenus Glad it could be of use :)

Certainly feel free to tag me if you run into any further issues.

PR #67, opened 2020-09-27
refactor: simplify touch event type check

web-padawan:
Updated `isTouch` with user defined type guard, which makes the code a bit cleaner by removing extra type cast.


omgovich:
Thanks @web-padawan ❤️

PR #71, opened 2020-10-02
feat: generate one file for types

jeetiss:
use `dts-bundle-generator` for types

this tool concatenates all types in one file and tree-shakes everything that not exported, but i'm not sure about breaking changes in this PR, and wanna write test for types!

@omgovich what do you think about that?


omgovich:
Hey @jeetiss. Thanks for your contribution.
The specs have failed. Probably because `tsconfig.microbundle.json` doesn't exist.

jeetiss:
@omgovich ye, i fix it

rschristian:
This generates an output that uses reference types rather than imports. This would completely break TS Preact support and force this library to have dependencies. If included in a TS application, that application cannot compile unless `@types/react` and `prop-types` are installed.

rschristian:
This would be a breaking change

omgovich:
Sounds bad. Should we close the PR then?

rschristian:
Maybe there's a similar alternative?

jeetiss:
@rschristian thanks for review!

let's dive deeper. Current types contain a lot of references to react. Is this a problem?
just one example `components/HslColorPicker.d.ts`:
Снимок экрана 2020-10-04 в 17 38 38

As I understand u can't use any react library without installing `@types/react` in typescript.

I skip reference to `prop-types` thanks for suggestion

rschristian:
I haven't been testing this lately, but somewhere along the ways it broke. Can get around with a `skipLibCheck` now though. 

> As I understand u can't use any react library without installing @types/react in typescript.

No, you can. 

@omgovich Not sure what you want to do here. It looks like this can be worked around by adding a field to the `tsconfig` file of all Preact TS apps that want to consume this library, so we could document that. However it might just be worth stating that Preact TS isn't fully supported. Can hand off the responsibility for that onto the users. I think until TS adds proper aliasing this will always be a headache trying to manage.

omgovich:
I like the idea to have only one d.ts file, but don't want to break compatibility accidentally.
So I would wait until microbundle handles that on their side (Ryan has created an issue in their GH repo).
Let's close this PR for now, but will keep in mind that we should try to flatten all declarations eventually.

jeetiss:
@omgovich please, show me issue at microbundle

rschristian:
I did not create an issue, I linked to a closed one that stated they weren't really interested in doing that as it serves no functional purpose.

rschristian:
Personally, while I get the appeal of one `.d.ts` file, I'd argue against it. It's added tooling to work around functionality that TS just doesn't support or agree with. Microbundle uses TSC to get those types files. 

While I do think it's cleaner, I'd personally weigh in against such functionality. Just another dependency and another thing that could go wrong with 0 benefits to the user experience.

omgovich:
Agree. Let's keep these things simple and easy-maintained.

PR #74, opened 2020-10-12
Fix pointer displaying in Safari 14

### LINKED ISSUE #73: The color is not updated after dragging the handle to the corner
CleanShot 2020-10-12 at 19 32 14

https://codesandbox.io/s/react-colorful-demo-u5vwp

omgovich:
Fixes #73 

Safari is a new IE...

Safari 14 doesn't work properly with `currentColor`, so I had to replace a pseudo-element that has `background-color: currentColor` with a regular `div` having `background-color`.

@rschristian Could you take a look?

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     {
       "path": "dist/index.module.js",
       "import": "{ RgbaStringColorPicker }",
-      "limit": "2 KB"
+      "limit": "2.1 KB"
     },
     {
       "path": "dist/index.module.js",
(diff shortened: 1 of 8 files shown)


rschristian:
Sure thing, I'll look through now

rschristian:
Strangely enough I can't reproduce the bug on iOS Safari. 

Your fix looks good to me tough.

omgovich:
Thanks, @rschristian!

P.S. I sent you PM on Twitter. Could you check when you have time? Need your help with an article)

omgovich:
Seems to me that it's only **desktop** Safari v14 issue.

PR #77, opened 2020-10-20
Round output values

### LINKED ISSUE #76: hsla editor lockup on quick update
Hi, Thanks very much for the component, react-color was indeed very heavy.

I have integrated it already and would be great if you can list it as one of the projects using react-colorful: https://component-controls.com/api/editors-coloreditor--overview

I seem to have an issue with fast updating of the control values with the hsla (and rgba) color picker (the hex color picker seems fine), any ideas what could be going on?

Here is a video, from the page linked above and the control locking up on fast moving around:

react-colorful

Also I resorted to converting rgba/hsla values since the HSLAStringColorPicker and rgba equivalent are not rounding the values, so I could end up with lets say 39.999999999 a one of the values. Is there an option for rounding the values (in my case I need to round to 0 decimals for the colr values and 2 decimals for the alpha value)

omgovich:
Closes #76

## Problem

image

All color models except HEX and RGB are awkward to use because output values have too many digits after the decimal point.

## What I did

The internal value is still super accurate (like `{ h: 1.23456789, s: 2.3456789, v: 3.4567, a: 0.9999999 }`).
Why? To keep the movement of pointers super smooth even when the picker size is 1000x1000.

This PR makes all output values rounded to the nearest integer.
Only alpha value has 2 digits to appear after the decimal point.

```js
hsl(200.4326236, 25.124125%, 32.121212%, 0.6666666) // will be impossible to get after merging this PR
hsl(200, 25%, 32%, 0.67) // that's what we going to return
```

@rschristian @molefrog Could I ask you to review the PR?

### DIFF
--- a/src/utils/round.ts
+++ b/src/utils/round.ts
@@ -1,0 +1,5 @@
+export const round = (number: number, digits = 0): number => {
+  // We wrap `toFixed` call with `Number()` to remove insignificant trailing zeros
+  // because `(1).toFixed(2) // 1.00`
+  return Number(number.toFixed(digits));
+};
--- a/src/utils/round.ts
+++ b/src/utils/round.ts
@@ -1,5 +1,3 @@
-export const round = (number: number, digits = 0): number => {
-  // We wrap `toFixed` call with `Number()` to remove insignificant trailing zeros
-  // because `(1).toFixed(2) // 1.00`
-  return Number(number.toFixed(digits));
+export const round = (number: number, digits = 0, base = Math.pow(10, digits)): number => {
+  return Math.round(base * number) / base;
 };
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@
 
 ## Features
 
-- **Small**: Just 1,8 KB gzipped (18 times lighter than **react-color**).
+- **Small**: Just 1,9 KB gzipped (17 times lighter than **react-color**).
 - **Tree-shakeable**: Only the parts you use will be imported into your app's bundle.
 - **Fast**: Built with hooks and functional components only.
 - **Bulletproof**: Written in strict TypeScript and covered by 40+ tests.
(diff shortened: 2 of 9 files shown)


omgovich:
New NPM/Node versions break specs (size-limit doesn't work there). Will try to fix that in a separate PR.
Ignore it for now. The PR is ready for review.

src/utils/round.ts
```diff
@@ -0,0 +1,5 @@
+export const round = (number: number, digits = 0): number => {
+  // We wrap `toFixed` call with `Number()` to remove insignificant trailing zeros
+  // because `(1).toFixed(2) // 1.00`
+  return Number(number.toFixed(digits));
```
molefrog, line 4: This doesn't seem very efficient. Converting a number to string back and forth means more memory allocations. Considering that the conversion usually happens during a UI interaction, this could result in worse performance. How about:

```js 
export const round = (number: number, digits = 0, base = Math.pow(10, digits)): number => {
  return Math.floor(base * number) / base
}
```

molefrog (approved):
Looks good!

src/utils/round.ts
omgovich: toFixed doesn't convert a number to a string. It converts the number to a number. Just moving the point and rounds a number.

omgovich: Oh shit, you're right. It returns a string...

molefrog: @omgovich Unfortunately, it does:
```js
typeof (123.234).toFixed(1)
"string"
```

I agree the method name is absolutely confusing 😄

molefrog: Btw, the `Math.pow` implementation isn't ideal: it'll break when `digits` reaches the number overflow. But that's fine unless we plan to use the method externally.

omgovich: image

Your solution is way faster, thx!

PR #83, opened 2020-10-22
refactor: Fixing for best Skypack package score

rschristian:
## Explanation

If you didn't know, [Skypack Discover launched](https://www.skypack.dev/blog/2020/10/introducing-skypack-discover/) and it aims to help with package search. To do this, it employs a scoring tool and [react-colorful's score is missing some low-hanging fruit](https://www.skypack.dev/view/react-colorful). This fixes those issues. 

## Changes
- Removed the `.npmignore` and moved back into `"files"`. I was the one who originally moved away from `"files"` as apparently `package.json` doesn't follow standard glob procedure and I couldn't find a way around that (until now). Skypack prefers `"files"` so why not fulfill that. No change in functionality.
- I also added an export map for the package. This just helps define what target uses what. Super simple addition since we already use Microbundle.


package.json
```diff
@@ -104,6 +112,14 @@
       "\\.css$": "identity-obj-proxy"
     }
   },
+  "files": [
+    "dist/*.{js,ts,css}",
+    "dist/components/*.ts",
+    "LICENSE",
+    "ACKNOWLEDGMENTS",
+    "README.md",
+    "package.json"
```
omgovich, line 121: BTW, according to the NPM's docs we probably could get rid of  `LICENSE`, `README.md`, and `package.json`. Could you check?
https://docs.npmjs.com/files/package.json

image

rschristian: We could. I mentioned this back on #59 too. I'd argue keeping them is a bit more explicit and clear, so it doesn't require any knowledge to know that they are always added. I can cut them out if you'd rather they weren't included though.

omgovich: I think you're right. Let keep them.

omgovich:
Noticed that the file count was 25 instead of 29 as before.
Got source maps back to the package.

rschristian:
Oh, crap, sorry.

omgovich (approved):
Thanks, @rschristian! You rock!

omgovich:
image

Issue #82, opened 2020-10-21
css modules


atanasster:
Hi, 

I am not using the css style imports and this makes it hard to match the generated class names, as they are set up with css modules.

Here is my uses case, you will notice that I ended up using some strange selectors (`div[role=slider]` and a few other places) to accommodate that some class names are not possible without using css modules.

https://github.com/ccontrols/component-controls/blob/9f68190dbd39f50a3bb522646262410b8bd7e37e/ui/editors/src/ColorEditor/ColorEditor.tsx#L83

I would suggest to just hard-code the class names prefixed with 'react-colorful' so they stay fixed

atanasster:
Just some more info on my issue, if it was not clear above:

Here is a div that didn't end up with a class name on my end: 
![grab106](https://user-images.githubusercontent.com/6075606/96754984-156b9800-13a0-11eb-8ca6-a1f6b6711fec.jpg)

Here is the same div from the official site: https://omgovich.github.io/react-colorful/ site (pointerFill class):
![grab107](https://user-images.githubusercontent.com/6075606/96755459-bfe3bb00-13a0-11eb-82b9-3e083faa939e.jpg)

You can play around with the class names without using css loaders on my end: https://component-controls.com/api/editors-coloreditor--overview

omgovich:
Hi, @atanasster!

- you don't have CSS-loader configured
- you don't want to configure one since your package is 100% CSS-in-JS
- you don't provide CSS-file with your package

Right?

omgovich:
I'm actually don't like that you have to copy our styled. 
We should find some way to import our stylesheet to your project.

We had a [discussion](https://github.com/omgovich/react-colorful/issues/72) about that with other maintainers, but haven't found a good solution yet.
But we'll try to help you somehow.

omgovich:
@molefrog I probably need your advice here.

atanasster:
Hi @omgovich ,

Yes I am 100% CSS-in-JS and do not have (or want to add) any css loaders. In this specific case I amusing theme-ui with emotion sx prop styling. 

I am not aware of any cross-style solution (that exports styles in css/e,otion/styled-components/etc.)

omgovich:
In styled-components it's possible to import and apply a stylesheet like that:
```js
import pickerStyles from 'react-colorful/dist/index.css' // if CSS-loader is configured to load .css as strings

const Element = styled.div`
  ${pickerStyles};
`
```

[...]

PR #84, opened 2020-10-22
Experiment: CSS modules → static class names

omgovich:
Relates to #82


atanasster (commented):
Looks great, maybe prefix the Static class names with ie ‘react_colorful__’?

I might be missing something but it seems class names like ‘preview’ could be clashing in a global namespace.

omgovich:
`preview` and `header` are our demo page styles. Don't worry about them. We don't publish them to NPM.

omgovich:
> Looks great, maybe prefix the Static class names with ie ‘react_colorful__’?

@atanasster Just tested this option. Makes the bundle size bigger.
Nothing bad to have a lot of `react-colorful` everywhere since gzip good in terms of this kind of compressions.

PR #87, opened 2020-10-26
CSS-in-JS experiments

omgovich:
The first (possible) step of our "CSS-in-JS" project.

Example 1:
```jsx
import { HexColorPicker, styles } from 'react-colorful'
import styled from '@emotion/styled'

const Container = styled('div')`
  ${styles};
`

const YourComponent = () => {
  return (
    <Container>
      <HexColorPicker color={color} onChange={setColor} />
    </Container>
  )
};
```

Example 2:
```jsx
import { HexColorPicker, styles } from 'react-colorful'
import { Global } from '@emotion/core'

const YourComponent = () => {
  return (
    <>
      <Global styles={styles} />
      <HexColorPicker color={color} onChange={setColor} />
    </>
  )
};
```

By exporting CSS as a string we could open an opportunity to create a recipes for most of CSS-in-JS libraries or even create our own wrappers around react-colorful for each CSS-in-JS solution.

@rschristian @molefrog Thought?

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,13 +21,17 @@
     "size": "npm run build && size-limit",
     "check-types": "tsc --noEmit true",
     "test": "jest tests --coverage",
-    "build": "del-cli 'dist/*' && microbundle build --entry src/index.ts --jsx React.createElement --name react-colorful --css-modules false --tsconfig tsconfig.build.json",
+    "build": "del-cli 'dist/*' && microbundle build --entry src/index.ts --jsx React.createElement --name react-colorful --css-modules false --tsconfig tsconfig.build.json && node scripts/inject-css-to-js-dist.js",
     "prepublishOnly": "npm run build",
     "start-demo": "parcel demo/src/index.html --out-dir demo/dist --open",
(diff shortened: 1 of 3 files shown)


rschristian:
If we wait for the cjs bug to be resolved with Microbundle we won't have to add more tooling

omgovich:
@rschristian No rush. This PR is an experiment/proof-of-concept/discussion.

omgovich:
I would like to understand the approach we're going to choose.

№1 = Merge this PR/approach and provide recipes (like the code above) for popular CSS-in-JS libraries. 

№2 = №1 + release more packages like `react-colorful-emotion` or `react-colorful-styled-components` for each library

№3 = №1 + release one package that will just inject styles when some of our components mount first time. 

№4 ?

omgovich:
We could try to use a Babel plugin (like [babel-plugin-import-require-as-string](https://github.com/Treverix/babel-plugin-import-require-as-string)), but most of these plugins are abandoned.

Can't find a good plugin that will also compress styles before injecting.

rschristian:
Or could try to fix the Microbundle issue ourselves. I'll certainly give it a shot, never looked into the source (or rollup for that matter) so can't say how well that'll go.

omgovich:
Migrated to #101

PR #94, opened 2020-11-25
change useLayoutEffect to useEffect

clifinger:
change useLayoutEffect to useEffect to be fully compatible with next.js


omgovich:
Hi @clifinger! 

We actually need `useLayoutEffect` here because `toggleDocumentEvents` should be triggered synchronously after user started dragging.

I'm going to create a Next.js playground for myself and think about a proper fix. 

Probably will go with solution like this:
https://github.com/reduxjs/react-redux/blob/8605088606403f6909597405c67db91d907db86e/src/components/connectAdvanced.js#L35-L41

omgovich:
@clifinger Thanks for your report! The problem is solved in `v4.4.3`.

PR #98, opened 2020-12-14
Migrate to PointerEvent

omgovich:


omgovich:
The bug is fixed via #99

PR #101, opened 2020-12-21
CSS-in-JS

omgovich:
A lot of projects having trouble installing **react-colorful** because they do not have any CSS-loader.
Also, it blocks many developers from including our library as part of their public component libraries since most of them are 100% CSS-in-JS.

For example, I wanted to add **react-colorful** to [react-three-gui](https://github.com/birkir/react-three-gui) but this project bundles via `tsdx` and do not support .css at all. 

Solution: move the styles of our picker to JS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@
 
 ## Features
 
-- **Small**: Just 1,9 KB gzipped (17 times lighter than **react-color**).
+- **Small**: Just 2,5 KB gzipped (14 times lighter than **react-color**).
 - **Tree-shakeable**: Only the parts you use will be imported into your app's bundle.
 - **Fast**: Built with hooks and functional components only.
 - **Bulletproof**: Written in strict TypeScript and covered by 40+ tests.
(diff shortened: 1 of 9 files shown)


omgovich:
@molefrog Can you take a look?

omgovich:
@rschristian I would be glad to hear your opinion as well ❤️

molefrog:
All looks good! Just a thought: if I decide to override the styles with my own ones, how will I be able to tell react-colorful to ignore the default styles? I mean I would have to deal with conflicting CSS in that case, right?

omgovich:
Right. But, at the same point, the component's styles are too complex to write from scratch.

molefrog:
@omgovich Yeah, you are right. I guess for the sake of backward-compatibility, the proposed solution is really good! 

The only downside is that once you upgrade, you're going to see an error that the styles are missing. But imo, with a good migration message in the README, this might be doable.

molefrog:
So I see two options here:
- Publish a new major release
- Don't remove the `styles.css` file yet in the next minor release, but make them redundant with a proper warning (will that be even possible?). And then remove the styles completely after everyone has migrated.

omgovich:
My plan is to release it as `v5`.

molefrog (approved):

rschristian (approved):
By all means, looks great to me.

PR #110, opened 2021-02-19
feat: Adding `getStyleSheet()` hook

### LINKED ISSUE #109: Feature: Better support SSR
I'm writing this PR right now, just thought I'd explain it a bit more in depth first in an issue. I find that helps separate comments about the feature from comments about the specific implementation.

temp

This comment left on your most recent post got me thinking about SSR, and a way we can improve it.

Firstly, we can provide a hook to the user to allow them to access the CSS string themselves. They can just inject it in the head of their doc if using SSR, for example. While this would be duplicated on load, that wouldn't be an issue.

But, to get around that, we could add an ID to the style element. So `useStyleSheet()` looks for a `<style id="__react-colorful">`, and only if it doesn't find one does it inject a new tag into the head. This way you not only support SSR, but make sure it's efficiently supported at the same time.

Should be quite slick and cover the use cases.

rschristian:
Closes #109

### DIFF
--- a/package.json
+++ b/package.json
@@ -113,6 +113,7 @@
   "files": [
     "dist/*.{js,ts,map}",
     "dist/components/*.ts",
+    "dist/hooks/useStyleSheet.d.ts",
     "LICENSE",
     "ACKNOWLEDGMENTS",
     "README.md",
--- a/src/hooks/useStyleSheet.ts
+++ b/src/hooks/useStyleSheet.ts
@@ -3,7 +3,7 @@ import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect";
 // Bundler is configured to load this as a processed minified CSS-string
 import styles from "../css/styles.css";
 
-const styleId = '__react-colorful';
+const styleId = "__react-colorful";
 let styleElement: HTMLStyleElement | undefined;
 
 /**
--- a/src/hooks/useStyleSheet.ts
+++ b/src/hooks/useStyleSheet.ts
@@ -3,6 +3,7 @@ import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect";
 // Bundler is configured to load this as a processed minified CSS-string
 import styles from "../css/styles.css";
 
+const styleId = '__react-colorful';
 let styleElement: HTMLStyleElement | undefined;
 
 /**
@@ -11,8 +12,18 @@ let styleElement: HTMLStyleElement | undefined;
 export const useStyleSheet = (): void => {
   useIsomorphicLayoutEffect(() => {
     if (typeof document !== "undefined" && !styleElement) {
+      if (document.getElementById(styleId)) {
+        styleElement = document.getElementById(styleId) as HTMLStyleElement;
+        return;
+      }
       styleElement = document.head.appendChild(document.createElement("style"));
+      styleElement.id = styleId;
       styleElement.innerHTML = styles;
     }
   }, []);
 };
+
+/**
+ * Allows CSS string to be accessible to the end user, for cases like SSR
+ */
+export const getStyleSheet = (): string => `<style id="${styleId}">${styles}</style>`;
--- a/src/index.ts
+++ b/src/index.ts
@@ -18,3 +18,6 @@ export { HexColorInput } from "./components/HexColorInput";
 
 // Color model types
 export { RgbColor, RgbaColor, HslColor, HslaColor, HsvColor, HsvaColor } from "./types";
+
+// CSS String
+export { getStyleSheet } from './hooks/useStyleSheet';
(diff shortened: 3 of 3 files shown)


omgovich:
Hi @rschristian! Thanks gor creating this PR!
Can you please describe how developers should work with the new method? I'm not sure I understand the way to connect that to SSR.

rschristian:
Sure, and apologies, meant to get a demo/recipe together for the ReadMe but was interrupted and forgot. 

Obviously it will depend on the end users' setup a bit, but this would allow server-side access to the CSS string + style tag. With this, you can do a simple `writeFileSync()` or whatever to inject the styles into the document before being sent to the client. The client, on hydration, would then check to see if the style tag already exists and only inject if it does not.

A slight perf improvement for those who want it or don't like the idea of a lib doing client-side injection like that.

I will try to find a demo somehow, not exactly sure what exists in this realm though besides like a GitHub gist. I don't think CodeSandbox or the like will really show off SSR properly, if at all. Not a big user of any of those, so maybe I'm wrong! Will look into it further.

omgovich:
Should we maybe provide a `<ColorPickerStyles />` component then?

```js
import Document, { Html, Head, Main, NextScript } from 'next/document'
import { ColorPickerStyles } from 'react-colorful'

class MyDocument extends Document {
  render() {
    return (
      <Html>
        <Head>
          <ColorPickerStyles />
        </Head>
        <body>
          <Main />
          <NextScript />
        </body>
      </Html>
    )
  }
}

export default MyDocument
```

omgovich:
Since `getStyleSheet` returns a string, I'm as a React developer have no idea what to do with it. From the framework's point of view, it's unsafe HTML, so there is no beautiful way to inject that during React-based SSR (e.g. Next)

rschristian:
For starters, some frameworks don't let you render what will became the head of the doc. Vite won't: https://github.com/vitejs/vite/blob/main/packages/playground/ssr-vue/prerender.js#L27 . I know Preact's tools won't either. You have to insert after the build is completed. And while, sure, you can use whatever `render-to-string` you'd want, really, this doesn't make much sense. You don't gain anything from it being a component. 

I've never used Next before, but I'm sure there's a way to inject a string.

omgovich:
I would close the PR if you don't mind. It seems to me that SSR is not important for a tiny color picker component.
I assume we can just ignore the problem.

rschristian:
Sure, no worries!

chrissantamaria:
👋 just wanted to chime in with my perspective here. I'm fairly new to the package so perhaps I don't have full context, but SSR is definitely a nice-to-have for my projects. Moving to `useStyleSheet` without any way of manually injecting styles makes that near impossible which is a bummer.

Beyond SSR, my (potentially uninformed) perspective is that react-colorful shouldn't impose its own CSS injection method on devs. One downside I've run into is that I can no longer process react-colorful styles with my PostCSS pipeline (since styles are baked into the JS bundle). A nice middle ground may have been making `useStyleSheet` opt-in, though I don't see any easy way of transitioning existing users.

Not trying to be argumentative and I respect the decision made, just wanted to share my thoughts. Thanks for the great package!

rschristian:
> Beyond SSR, my (potentially uninformed) perspective is that react-colorful shouldn't impose its own CSS injection method on devs. 

So we talked about this quite a bit, and I'd say over quite a period of time too.

The issues that we looked to solve here were a) keeping size down, b) keeping the authoring experience reasonable (no crazy build steps), and c) allowing use of this library to be as simple as possible. Now Microbundle (the build tool we use) follows the JS spec, which means all assets need to be imported by the end user separately in accordance with their build tools. I don't think anyone is a huge fan of that, as it's quite messy, though I totally agree with following the spec in that regard (admittedly I am biased as I help maintain it).

This wasn't ever about "imposing" anything, but trying to offer the simplest use scenario for most people.

That being said, maybe options wouldn't hurt. @omgovich What do you think about maybe offering this PR as well as shipping the authored CSS with an `useStyleSheet()` opt-out? This would give SSR users an easy hook in and shipping the CSS file itself would allow users to process it however they wanted. I do think any processing of the authored CSS by the user means we'd have to allow them to disable `useStyleSheet()`, as it'll likely conflict otherwise, but it could be a zero-cost change to existing users.

JS size increase would only be marginal, likely. Might be able to pick up more users that way too.

chrissantamaria:
> This wasn't ever about "imposing" anything, but trying to offer the simplest use scenario for most people.

I totally agree! Impose wasn't the right word - just meant a decision is being made on behalf of devs without any alternatives.

Happy to help with any work around these changes, even if just examples for frameworks like Next or Gatsby.

omgovich:
Hey, guys! Thanks for being involved!
Could you please share your thoughts about API update?

As far as I understand you did offer something like the code below?

```js
import { HexColorPicker, styles } from 'react-colorful'

// do something with `styles`

<HexColorPicker unstyled />
```

rschristian:
So with this PR, the `getStyleSheet()` hook, something like that, though we wouldn't have to alter the component API. Instead, the existing `useStyleSheet()` would check upon DOM hydration to see if the style tag already existed. 

But if the user wants to import our raw, as-authored CSS (which they may have a use case for) then I think something like that `unstyled` prop would be necessary.

PR #119, opened 2021-03-10
HTML attributes support

omgovich:
I was working on some demos recently and noticed that it's not possible to add `id` or another HTML attribute to the root div. Also the picker does not support native DOM events that a regular `div` does (e.g. `onMouseEnter`).

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,3 +1,5 @@
+import React from "react";
+
 export interface RgbColor {
   r: number;
   g: number;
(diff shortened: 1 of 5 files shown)


omgovich:
@rschristian 👋 Calling for our TS guru =)

src/types.ts
```diff
@@ -39,8 +41,12 @@ export interface ColorModel<T extends AnyColor> {
   equal: (first: T, second: T) => boolean;
 }
 
-export interface ColorPickerBaseProps<T extends AnyColor> {
-  className: string;
+type ColorPickerHTMLAttributes = Omit<
+  React.HTMLAttributes<HTMLDivElement>,
+  "color" | "onChange" | "onChangeCapture"
```
rschristian, line 46: What's the reason for `"color"`? As far as I know no such attribute exists, so I can't imagine there's even any type collision there?

omgovich: I was super confused as well, but it's present in React's default types:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1785

omgovich: image

omgovich: I assume it's related to super-legacy stuff like `<font color="red">text</font>`.

rschristian: Huh, weird. Even in super old legacy stuff that shouldn't apply to an HTMLDivElement? Gotta question what those React folks are doing some days. 

But fair, makes sense

rschristian (approved):
Certainly looks all good to me, though I do now feel a bit more sympathy for the React developer who has to deal with this nonsense 😅

molefrog (approved):
This looks good!

PR #121, opened 2021-03-25
CSP: Support `nonce`

### LINKED ISSUE #120: Added styles do not conform to any CSP rules.
Hey I am using this cool component in my react project, but I am having an issue with the Content Security Policy.

Although I am doing client side rendering with `makeStyles `from material-ui, I use the nonce because I am building an app with electron and want to secure it as much as possible.

Right now I am using a nonce to only allow self styles and styles with a specific nonce. The `useStyles` hook created by `makeStyles` does add the nonce to the style because it uses [JSS](https://cssinjs.org/)

The problem is that styles in react-colorful are applied to the html without [JSS](https://cssinjs.org/) so no nonce is added.

Possibly related topics are: #109 #110

Any idea on how to solve the problem?

omgovich:
Closes #120

### DIFF
--- a/src/hooks/useStyleSheet.ts
+++ b/src/hooks/useStyleSheet.ts
@@ -1,4 +1,5 @@
 import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect";
+import { getNonce } from "../utils/nonce";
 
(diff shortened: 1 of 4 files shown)


molefrog (approved):
This looks good!

PR #127, opened 2021-04-21
fix: Correcting ESM usage with Node

### LINKED ISSUE #125: SyntaxError: Cannot use import statement outside a module
I get the following issue when trying to use react-colorful in my NextJS app:

```
SyntaxError: Cannot use import statement outside a module
 module.exports = require("react-colorful");
```

```
import { HexColorPicker } from "react-colorful";
<HexColorPicker color={color} onChange={setColor} />
```

Anyone know why?

rschristian:
I believe this should solve #125. 

Use in a `"type": "module"` (i.e., Node native ESM) app is certainly broken right now anyhow, so this should at least fix that problem, even if it doesn't solve #125. Will need to wait for comments from the user, as the error they see should be different, though that might just a Node version difference. Not yet sure.


package.json
```diff
@@ -5,14 +5,13 @@
   "source": "src/index.ts",
   "main": "dist/index.js",
   "module": "dist/index.module.js",
-  "esmodule": "dist/index.esmodule.js",
```
rschristian, line 8: This was suggested by Microbundle but has been replaced entirely with `"exports"` as of Node 14 I believe. I think a few tools support it, but it's not worth having really.

We could (should?) make the `"browser"` key the consumer of the "modern" output.

omgovich:
Hi @rschristian. Thanks for your work!
Question: can't we drop `.module.js` and use `.mjs` everywhere?
Same way as I do in [Colord](https://github.com/omgovich/colord/blob/master/package.json)

package.json
omgovich: Not sure about that 🤔 Anyway, we have to postpone that's until the next major release.

rschristian:
No, `.mjs` is not supported by all tools / environments at the moment.

package.json
rschristian: Any build tool would fall back to plain ESM, but you're right, I suppose if someone's using a direct URL via a CDN or something then that's no good. My bad. Will revert.

omgovich:
Sounds like you're right. I'll dig into that and double-check everything tomorrow.
I'm afraid to break something since the library is quite popular already :)

P.S. What is "postbuild" BTW?) Can't find that in both NPM and Microbundle docs.
Don't you think that we can replace the new .js-file with two simple `cp` calls inside package.json/scripts?

rschristian:
Just from the last few days at work, I ran into older Create-React-App versions not having `.mjs` as file extensions to handle, so how Webpack would deal with them could get whacky. Then some tools just flat out don't recognize it, as it only got popular since Node 14, which is still quite new to be fair. 

"postbuild" isn't a built-in, but Node supports "pre" and "post" as prefixes to matching commands. So "preX" and "postX" both work so long as "X" exists. [Docs](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts). 

CP does work, however, `cp` is a Unix thing; it doesn't exist in Windows. Don't know what your policy there is for support. Up to you.

omgovich:
> So "preX" and "postX" both work so long as "X" exists. [Docs](https://docs.npmjs.com/cli/v6/using-npm/scripts#pre--post-scripts).

Damn, didn't know that! Thanks for sharing!

> `cp` does work, however, it is a Unix thing; it doesn't exist in Windows. Don't know what your policy there is for support. Up to you. If someone wants to try out the latest `react-colorful` by building from source and linking, they might have to manually copy those files if they're on Windows as `cp` won't be recognized as a valid command.

What if we use something like https://github.com/sindresorhus/cpy-cli to perform copying and renaming inside of package.json?

rschristian:
Certainly up to you. I went the safe, dep-free route for a first pass but could add that in if it's preferable.

That being said, it's super late here and I got to get to bed. I won't have that fixed for another 10-12 hours probably, so if you want to get this in before then feel free to make that change on on this branch

rschristian:
@omgovich Pushed with copy utility. Ended up using NCP. Does the job just as well, but a bit lighter on install size according to Packagephobia. Not a big deal either way though.

[ncp](https://packagephobia.com/result?p=ncp), [cpy-cli](https://packagephobia.com/result?p=cpy-cli)

package.json
```diff
@@ -23,6 +23,7 @@
     "check-types": "tsc --noEmit true",
     "test": "jest tests --coverage",
     "build": "del-cli 'dist/*' && microbundle build --entry src/index.ts --jsx React.createElement --name react-colorful --css inline --tsconfig tsconfig.build.json",
+    "postbuild": "ncp dist/index.module.js dist/index.mjs && ncp dist/index.module.js.map dist/index.mjs.map",
```
omgovich, line 26: Thanks! 

The only thing that bothers me now is that `ncp` has no updates for 6 years =)
I would go with `cpy-cli`. The API is more complicated, but the library is more "fresh" and the author maintains it:
```suggestion
    "postbuild": "cpy dist/index.module.js dist --rename=index.mjs && cpy dist/index.module.js.map dist --rename=index.mjs.map",
```

rschristian: Tends to be because such things are super stable. Windows isn't going to change it's copy method over tomorrow. 

But alright, will swap it out.

rschristian:
I did unlink this from #125. While what this PR solves is _an issue_, I'm not confident it's _the issue_ that user is seeing. The error message posted is odd, but I can't tell if they've just copy/pasted a weird section of the full message or not.

rschristian:
Crap, sorry. Didn't mean to do that!

My version autofill can be a bit slow, so I usually fill "latest" first and come back when it'll suggest to me the right version.

omgovich:
No problemo)

omgovich:
Should we test it somehow, how do you think? Maybe release a beta.
It feels like we can just release it since Preact uses the same approach.

omgovich:
P.S. "exports" map is super PITA.
The modern approach expects libraries to have "exports" field, but most of the packages I saw just ignore that and don't have problems we do at all =)

rschristian:
How many beta downloads do you normally get? Testing it locally seems to be all good, both with Node & bundlers, but we can't test for every tool obviously. If you'd get enough beta testers that could be good, else, it's a quick revert if some odd tool breaks over this (not that I imagine any will)

Exports is definitely a pain, lol. But it only came out in Node v12.7.0, so less than two years ago. Now that Node 10 is dead ESM and exports adoption is going to pick up. Node 12 is the minimum supported version.

rschristian:
Will probably become easier when more people just switch fully to ESM. If you didn't know, [sindresorhus is planning on moving all of his to ESM](https://github.com/sindresorhus/meta/discussions/15), and that's `"type": "module"`, not `.mjs`. I imagine that'll help drive adoption.

omgovich:
Sounds great. I tried to make Colord ESM first and got an issue in the first week =))
https://github.com/omgovich/colord/issues/15

omgovich:
Works with Next.js and Parcel.

omgovich:
Works with Create-React-App.

I think we're good to go.
Thanks, @rschristian!

omgovich (approved):

PR #131, opened 2021-05-07
docs: Adding WMR to Preact usage section

rschristian:
With WMR being used a bit more now I figured it makes sense to add, just in case some user of yours reads that and thinks it's not supported.

And it does work flawlessly out of the box, which is great 😁


omgovich:
Agree! Thanks @rschristian!

PR #135, opened 2021-05-24
Feat: string types

rschristian:
Not ready for merge before TS 4.3 finalizes (using the RC now) but template literal types are now really, really good and easy to use!

```
type RgbStringColor = `rgb(${number}, ${number}, ${number})`;

const color: RgbStringColor = "rgb(a, 0, 0)";
```

``[tsserver 2322] [E] Type '"rgb(a, 0, 0)"' is not assignable to type '`rgb(${number}, ${number}, ${number})`'.``

IMO, that's a super easy error to understand and helps a lot. 

Now, unfortunately we can't narrow down the number ranges (like 0-255) and hex strings are pretty much out of the question due to union type complexities (6 digit hex creates a union type of ~16.7 million possibilities, which the TS language server is very much not a fan of). ~~We could try something like `type HexDigit = number | string`, but I'm not sure how useful that would be. Maybe though, as it does narrow down the type somewhat.~~ Edit: Of course that won't work as you can't set fixed string lengths. Hex is a no-go.

At the very least, this can catch common typos in the string types, even if it's not as accurate as wished for. TS will immediately inform users they accidentally typed `rbg(0, 0, 0);` for example; they won't need to wait for the runtime regex to know of an error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "react-colorful",
-  "version": "5.1.3",
+  "version": "5.2.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -12296,6 +12296,12 @@
           "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.1.0.tgz",
           "integrity": "sha512-hcVC3wYEziELGGmEEXue7D75zbwIIVUMWAVbHItGPx0ziyXxrOMQx4rQEVEV45Ut/1IotuEvwqPopzIOkDMf0A==",
           "dev": true
+        },
+        "typescript": {
+          "version": "4.2.4",
+          "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.2.4.tgz",
+          "integrity": "sha512-V+evlYHZnQkaz8TRBuxTA92yZBPotr5H+WhQ7bD3hZUndx5tGOa1fuCgeSjxAzM1RiN5IzvadIXTVefuuwZCRg==",
+          "dev": true
         }
       }
     },
@@ -17810,9 +17816,9 @@
       }
     },
     "typescript": {
-      "version": "4.2.3",
-      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.2.3.tgz",
-      "integrity": "sha512-qOcYwxaByStAWrBf4x0fibwZvMRG+r4cQoTjbPtUlrWjBHbmCAww1i448U0GJ+3cNNEtebDteo/cHOR3xJ4wEw==",
+      "version": "4.3.1-rc",
+      "resolved": "https://registry.npmjs.org/typescript/-/typescript-4.3.1-rc.tgz",
+      "integrity": "sha512-L3uJ0gcntaRaKni9aV2amYB+pCDVodKe/B5+IREyvtKGsDOF7cYjchHb/B894skqkgD52ykRuWatIZMqEsHIqA==",
       "dev": true
     },
     "unbox-primitive": {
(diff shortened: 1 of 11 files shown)


web-padawan:
Hi, maintainer of `vanilla-colorful` here 🙂 

This is something I was thinking about but didn't know about the improvement coming to TypeScript 4.3.
There are some questions that I would like to clarify related to implementing this feature:

1. Does this require users to upgrade to TS 4.3 in order to consume the component? I checked the [downlevel-dts](https://github.com/sandersn/downlevel-dts) project and they don't seem to support this feature yet. I'm not sure whether it is possible to downlevel these types at all.
2. What about supporting strings without spaces e.g. `hsl(0,0,100)` or modern color notations introduced by #133? 

I've tried to implement optional space using the following approach:

```ts
type AnySpace = '' | ' ';

export type RgbString = `rgb(${number},${AnySpace}${number},${AnySpace}${number})`;
```

But this makes the resulting type a bit long:

```
`rgb(${number},${number},${number})` | `rgb(${number},${number}, ${number})` | `rgb(${number}, ${number},${number})` | `rgb(${number}, ${number}, ${number})`
```

rschristian:
@web-padawan

1) This doesn't require TS 4.3, no. 4.3 just includes enough QoL benefits to finally make it worth having. For example, you used to need to manually cast return types, else they'd default to `string`. 

```
function bar(s: string): `hello ${string}` {
    // Previously an error, now works!
    return `hello ${s}`;
}
```

It doesn't require TS 4.3, but does 4.1, as that's when template types were added. If a user is using something older, the type will be `any`, which is pretty much just as uncontrolled as `string`. Can't down level this besides converting any template literal type back into a string.

2) Yes, this would require types for all those variants if you want to support them, and is why I think #133 was a bad idea. Increases the complexity by a lot allowing users to toss unfiltered formats at this lib. 

Alternatively, you could just tell users if they want to use those exotic types to toss in a `@ts-ignore`. The TS language server cannot handle complex unions which you'd quickly be limited by if you wanted to support all of those formats. Perfectly acceptable to tell the few to use ts-ignore so the rest get better hints, IMO.

omgovich:
Hi @rschristian! Pretty cool feature. I was trying to implement that to Colord, but failed since RGB[A] has different syntaxes and I'm going to add their support to react-colorful soon as well.
```
rgba(128,128,128,.5)
rgba(128, 128, 128, .5)
rgba(128, 128, 128, 0.5)
rgba(128, 128, 128, 50%)
rgba(50%, 50%, 50%, 50%)
rgba(128 128 128 / 0.5)
rgba(128 128 128 / 50%)
rgba(50% 50% 50% / .5)
rgba(50% 50% 50% / 50%)
rgb(255, 0, 153.6, 1)
rgb(1e2, .5e1, .5e0, +.25e2%)
```

I think it almost impossible to type all of these.

rschristian:
Fair enough, though I think trying to support all of those is a mistake. You're increasing the runtime complexity and bundle size by a lot and not really gaining much in return. 

Feel free to close this out though, I'm not invested or anything. Wanted to play with the new types and thought this would make a good playground.

web-padawan:
I think this is a case where we need to care about the developer experience and make a decision.

1. Support for modern color notations is a nice feature, but it adds complexity.
2. Not having to care about all possible colors makes it easier to add types but might be unexpected by some users.

IMO the component could be updated to make modern color support opt-in, e.g. by creating a different set of entrypoints.

Then the "strict" entrypoints (old style) could have the new types, and the "loose" entrypoints would support all the possible color notations, while changing the type to `string` instead. But implementing this would increase complexity, too.

omgovich:
I'm 100% sure that supporting valid RGB and HSL syntaxes is important. It's not difficult but has a pretty high value for developers.

At the same time, using types like `rgb(${number},${number},${number})` will probably only prevent people from using the library since (I'm pretty sure) the projects don't ready to work with the types of this kind. 100% of the projects have `string` type for their colors and we'll force them to find workarounds to define their colors in our way.

PR #150, opened 2021-07-31
Use pointer events instead of mouse and touch events. Support multi touch

xnimorz:
Hey! Thanks for the library ;)

I was looking at the library after our Twitter discussion: https://twitter.com/Omgovich/status/1421136814954385418

## Overview

This pull request introduces several changes:

### Removed MouseEvent and TouchEvent in favor of PointerEvent. 

It has several advantages:

1) Smaller library size (about 170 bytes):

|Before|After|
|---|---|
| image | image |
| image | image |

2) We don't have browser hacks in code to preventDefault behavior and etc.

3) We don't need to resolve conflicts between touches and clicks. 

### Improved library performance 

1) I removed `useState`, which reduces component re-renders, 
2) got rid of excess `useEffects`  along with `useCallbacks` which also optimizes comparisons and overall performance

### Multi touch support

(sorry, didn't find a better way to attach the screencasts)

|Before|After|
|---|---|
|https://user-images.githubusercontent.com/1178825/127752546-d7169520-c6d8-4495-9ab4-5af16505edd5.MP4|https://user-images.githubusercontent.com/1178825/127752561-12864f6e-7cad-4a56-b3de-dd6b069eb568.MP4|

### Unit tests support

1) Removed unit test which no more needed 🙈

2) Added `PointerEvents` support

## What about browsers support?

It should work in all browsers including IE11: (and partly IE10): https://caniuse.com/pointer
image

Some screencasts:
Chrome: https://user-images.githubusercontent.com/1178825/127753055-25ba0bb7-8dd0-481d-8776-4f94bd0898a9.mp4

Firefox: https://user-images.githubusercontent.com/1178825/127752834-2da54a31-7c57-47a6-bfc8-3be510af0ace.mov 

IPhone safari: https://user-images.githubusercontent.com/1178825/127752561-12864f6e-7cad-4a56-b3de-dd6b069eb568.MP4

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,9 @@
       "\\.js$": "jest-esm-jsx-transform",
       "^.+\\.tsx?$": "ts-jest"
     },
+    "setupFilesAfterEnv": [
+      "./tests/setup.js"
+    ],
     "moduleNameMapper": {
       "\\.css$": "<rootDir>/tests/__mocks__/styles.css.mock.ts"
     }
--- a/src/components/common/Interactive.tsx
+++ b/src/components/common/Interactive.tsx
@@ -1,6 +1,5 @@
-import React, { useState, useRef, useCallback } from "react";
+import React, { useRef, useCallback } from "react";
 
-import { useIsomorphicLayoutEffect } from "../../hooks/useIsomorphicLayoutEffect";
 import { useEventCallback } from "../../hooks/useEventCallback";
 import { clamp } from "../../utils/clamp";
 
@@ -9,27 +8,14 @@ export interface Interaction {
   top: number;
 }
 
-// Check if an event was triggered by touch
-const isTouch = (event: MouseEvent | TouchEvent): event is TouchEvent => "touches" in event;
-
 // Returns a relative position of the pointer inside the node's bounding box
-const getRelativePosition = (node: HTMLDivElement, event: MouseEvent | TouchEvent): Interaction => {
+const getRelativePosition = (node: HTMLDivElement, event: PointerEvent): Interaction => {
   const rect = node.getBoundingClientRect();
 
-  // Get user's pointer position from `touches` array if it's a `TouchEvent`
-  const pointer = isTouch(event) ? event.touches[0] : (event as MouseEvent);
-
   return {
-    left: clamp((pointer.pageX - (rect.left + window.pageXOffset)) / rect.width),
-    top: clamp((pointer.pageY - (rect.top + window.pageYOffset)) / rect.height),
+    left: clamp((event.pageX - (rect.left + window.pageXOffset)) / rect.width),
+    top: clamp((event.pageY - (rect.top + window.pageYOffset)) / rect.height),
   };
-};
-
-// Browsers introduced an intervention, making touch events passive by default.
-// This workaround removes `preventDefault` call from the touch handlers.
-// https://github.com/facebook/react/issues/19651
-const preventDefaultMove = (event: MouseEvent | TouchEvent): void => {
-  !isTouch(event) && event.preventDefault();
 };
 
 interface Props {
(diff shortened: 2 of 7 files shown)


src/components/common/Interactive.tsx
```diff
@@ -40,55 +26,35 @@ interface Props {
...
-      // If user moves the pointer outside of the window or iframe bounds and release it there,
-      // `mouseup`/`touchend` won't be fired. In order to stop the picker from following the cursor
-      // after the user has moved the mouse/finger back to the document, we check `event.buttons`
-      // and `event.touches`. It allows us to detect that the user is just moving his pointer
-      // without pressing it down
-      const isDown = isTouch(event) ? event.touches.length > 0 : event.buttons > 0;
-
-      if (isDown && container.current) {
+    (event: PointerEvent) => {      
+      if (container.current) {
```
xnimorz, line 34: Hey TS ;)

src/components/common/Interactive.tsx
```diff
@@ -40,55 +26,35 @@ interface Props {
...
+        return;
+      }
 
-      // Prevent text selection
-      preventDefaultMove(nativeEvent);
-
-      // Interrupt "mousedown" call on mobiles
-      if (!isValid(nativeEvent) || !el) return;
+      // testing-library doesn't have setPointerCapture implemented
+      el.setPointerCapture && el.setPointerCapture(nativeEvent.pointerId);
```
xnimorz, line 49: This simple line makes:
1) Multi touch support
2) All subsequent pointermove and pointerup events will be captured by this particular element, hence if the pointer is released outside the browser it still will be captured (see screencasts section in description)

More info: https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture

src/components/common/Interactive.tsx
```diff
@@ -40,55 +26,35 @@ interface Props {
 
 const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
   const container = useRef<HTMLDivElement>(null);
-  const hasTouched = useRef(false);
-  const [isDragging, setDragging] = useState(false);
```
xnimorz, line 44: Don't need it anymore, which improves the performance

src/components/common/Interactive.tsx
```diff
@@ -40,55 +26,35 @@ interface Props {
...
   const onMoveCallback = useEventCallback<Interaction>(onMove);
   const onKeyCallback = useEventCallback<Interaction>(onKey);
 
-  // Prevent mobile browsers from handling mouse events (conflicting with touch ones).
-  // If we detected a touch interaction before, we prefer reacting to touch events only.
-  const isValid = (event: MouseEvent | TouchEvent): boolean => {
-    if (hasTouched.current && !isTouch(event)) return false;
-    if (!hasTouched.current) hasTouched.current = isTouch(event);
-    return true;
-  };
```
xnimorz, line 54: Fulfilled by setPointerCapture fn

src/components/common/Interactive.tsx
```diff
@@ -40,55 +26,35 @@ interface Props {
...
-
   const handleMove = useCallback(
-    (event: MouseEvent | TouchEvent) => {
-      // Prevent text selection
-      preventDefaultMove(event);
```
xnimorz, line 59: Don't need it anymore, as there is no intervention for pointer events (as I know)

src/components/common/Interactive.tsx
```diff
@@ -40,55 +26,35 @@ interface Props {
...
+      el.setPointerCapture && el.setPointerCapture(nativeEvent.pointerId);
 
       // The node/ref must actually exist when user start an interaction.
       // We won't suppress the ESLint warning though, as it should probably be something to be aware of.
       el.focus();
       onMoveCallback(getRelativePosition(el, nativeEvent));
-      setDragging(true);
+
+      el.addEventListener("pointermove", handleMove);
+      el.addEventListener("pointerup", handleMoveEnd);
```
xnimorz, line 57: Events are attached to the element 'cause it captures all subsequent calls

src/hooks/useEventCallback.ts
```diff
@@ -1,14 +1,15 @@
-import { useRef, useEffect, useCallback } from "react";
+import { useRef } from "react";
 
+function noop() {}
 // Saves incoming handler to the ref in order to avoid "useCallback hell"
 function useEventCallback<T>(handler?: (value: T) => void): (value: T) => void {
   const callbackRef = useRef(handler);
 
-  useEffect(() => {
```
xnimorz, line 7: There is no need in pause, when a new function was passed, it doesn't have any async side effects, therefore removing useEffect makes the performance better

src/hooks/useEventCallback.ts
```diff
@@ -1,14 +1,15 @@
...
+  if (callbackRef.current !== handler) {
     callbackRef.current = handler;
-  });
+  }
 
-  return useCallback((value: T) => callbackRef.current && callbackRef.current(value), []);
+  return callbackRef.current || noop;
```
xnimorz, line 12: noop is to support backward compatibility

tests/components.test.js
```diff
@@ -148,30 +137,12 @@ it("Changes alpha channel value after an interaction", async () => {
...
-  fireEvent.touchStart(hue, { touches: [{ pageX: 0, pageY: 0, bubbles: true }] }); // 1
-  fireEvent.touchMove(hue, { touches: [{ pageX: 55, pageY: 0, bubbles: true }] }); // 2
-
-  // Should be skipped
-  fireEvent(hue, new FakeMouseEvent("mousedown", { pageX: 35, pageY: 0 })); // 3
-  fireEvent(hue, new FakeMouseEvent("mousemove", { pageX: 105, pageY: 0 })); // 4
-
-  expect(handleChange).toHaveReturnedTimes(2);
-  expect(handleChange).toHaveReturnedWith("hsl(180, 0%, 0%)");
-});
```
xnimorz, line 173: Pointer events make it unified

tests/components.test.js
```diff
@@ -91,54 +97,37 @@ it("Doesn't call `onChange` when user changes a hue of a grayscale color", () =>
...
-
-  // User presses and moves the cursor inside the window
-  fireEvent(saturation, new FakeMouseEvent("mousedown", { pageX: 20, pageY: 10 })); // 1
-  fireEvent(saturation, new FakeMouseEvent("mousemove", { pageX: 10, pageY: 10 })); // 2
-  // User releases the mouse button outside of the document bounds with no `mouseup` event fired
-  // User moves the cursor back to the document with no button pressed
-  fireEvent(saturation, new FakeMouseEvent("mousemove", { pageX: 1, pageY: 50, buttons: 0 })); // 3
-
-  expect(handleChange).toHaveReturnedTimes(2); // the last `mousemove` has to be ignored
-});
```
xnimorz, line 140: fulfilled by using pointerevents

xnimorz:
Added separate commit, which adds TouchEvent support only for browsers that don't support PointerEvents. 

The size gets bigger than after the first commit, but it is still smaller than the master state:
image

Also, the second commit removes all `useCallbacks` from the `Interactive.tsx` and makes some changes for unit tests setup (because ts-jest has big problems with mocking modules)

The only problem I have atm (didn't figure out it yet) is that `Captures arrow keys only` gets failed (I'll check it tomorrow)

package.json
```diff
@@ -124,6 +124,9 @@
       "\\.js$": "jest-esm-jsx-transform",
       "^.+\\.tsx?$": "ts-jest"
     },
+    "setupFilesAfterEnv": [
+      "./tests/setup.js"
```
xnimorz, line 128: https://github.com/facebook/jest/issues/2582 and it's not resolved (only workarounds)

src/hooks/useEventCallback.ts
omgovich: Smart!

tests/components.test.js
omgovich: Can we emulate the same test case (pointer outside of the window) with `PointerEvent`?

src/components/common/Interactive.tsx
omgovich: Nice!

src/hooks/useEventCallback.ts
omgovich: @jeetiss check this out. You might find this approach interesting.

omgovich:
Hey @xnimorz. I'm still not sure about migrating to `PointerEvent`. I try to support as many browsers as possible.

I used to drop support of browsers that have less than 0.5% of the marker (the same rule as **browserslist**'s `defaults` use). But there are a lot of browsers that don't support `PointerEvent` and have more than 0.5%.
- Firefox <=52 — `0.51%` (needs `MouseEvent`)
- Chrome <=54 — `0.6%` (needs `MouseEvent`)
- macOS Safari <=12.1 — `~0.5%` (needs `MouseEvent`)
- iOS Safari <=12.4 — `0.9%` (you covered that by the fallback to `TouchEvent`)

xnimorz:
I'll update the diff this week later and return mouse events than

xnimorz:
I'm going to publish 2 separate diff

PR #151, opened 2021-08-07
Optimize hooks, reduce library size and get rid of edge case errors and memory leaks

xnimorz:
This diff makes the runtime faster and the library size smaller

|Before|After|
|---|---|
| image | image |
| image | image |

As we get rid of `useEffect` calls, it saves time in runtime. 

It also prevents us from the memory leak and potential errors

The behaviour isn't changed:

https://user-images.githubusercontent.com/1178825/128600229-9f30f0ae-5cf6-4c31-ae57-62cb17e3bdce.mov

### DIFF
--- a/src/components/common/Interactive.tsx
+++ b/src/components/common/Interactive.tsx
@@ -1,4 +1,4 @@
-import React, { useRef, useMemo } from "react";
+import React, { useRef, useMemo, useEffect } from "react";
 
 import { useEventCallback } from "../../hooks/useEventCallback";
 import { clamp } from "../../utils/clamp";
@@ -47,7 +47,7 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
   const onKeyCallback = useEventCallback<Interaction>(onKey);
   const hasTouch = useRef(false);
 
-  const [handleMoveStart, handleKeyDown] = useMemo(() => {
+  const [handleMoveStart, handleKeyDown, toggleDocumentEvents] = useMemo(() => {
     const handleMoveStart = ({ nativeEvent }: React.MouseEvent | React.TouchEvent) => {
       const el = container.current;
       if (!el) {
@@ -103,7 +103,7 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
       });
     };
 
-    function toggleDocumentEvents(state: boolean) {
+    function toggleDocumentEvents(state?: boolean) {
       const touch = hasTouch.current;
       // add or remove additional pointer event listeners
       const toggleEvent = state ? self.addEventListener : self.removeEventListener;
@@ -111,8 +111,9 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
       toggleEvent(touch ? "touchend" : "mouseup", handleMoveEnd);
     }
 
-    return [handleMoveStart, handleKeyDown];
+    return [handleMoveStart, handleKeyDown, toggleDocumentEvents];
   }, []);
+  useEffect(() => toggleDocumentEvents, []);
 
   return (
     <div
--- a/src/components/common/Interactive.tsx
+++ b/src/components/common/Interactive.tsx
@@ -1,9 +1,7 @@
-import React, { useState, useRef, useCallback } from "react";
+import React, { useRef, useMemo } from "react";
 
-import { useIsomorphicLayoutEffect } from "../../hooks/useIsomorphicLayoutEffect";
 import { useEventCallback } from "../../hooks/useEventCallback";
 import { clamp } from "../../utils/clamp";
-
 export interface Interaction {
   left: number;
   top: number;
@@ -32,6 +30,11 @@ const preventDefaultMove = (event: MouseEvent | TouchEvent): void => {
   !isTouch(event) && event.preventDefault();
 };
 
+// Prevent mobile browsers from handling mouse events (conflicting with touch ones).
+// If we detected a touch interaction before, we prefer reacting to touch events only.
+const isInvalid = (event: MouseEvent | TouchEvent, hasTouch: boolean): boolean =>
+  hasTouch && !isTouch(event);
+
 interface Props {
   onMove: (interaction: Interaction) => void;
   onKey: (offset: Interaction) => void;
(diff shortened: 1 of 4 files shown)


src/components/common/Interactive.tsx
```diff
@@ -32,6 +30,11 @@ const preventDefaultMove = (event: MouseEvent | TouchEvent): void => {
   !isTouch(event) && event.preventDefault();
 };
 
+// Prevent mobile browsers from handling mouse events (conflicting with touch ones).
+// If we detected a touch interaction before, we prefer reacting to touch events only.
+const isInvalid = (event: MouseEvent | TouchEvent, hasTouch: boolean): boolean =>
```
xnimorz, line 35: The only usage of isValid is in `!isValid` if clause.

src/components/common/Interactive.tsx
```diff
@@ -40,21 +43,31 @@ interface Props {
...
+        return;
+      }
 
-  const handleMove = useCallback(
-    (event: MouseEvent | TouchEvent) => {
+      // Prevent text selection
+      preventDefaultMove(nativeEvent);
+
+      if (isInvalid(nativeEvent, hasTouch.current) || !el) return;
+      hasTouch.current = isTouch(nativeEvent);
```
xnimorz, line 61: I moved `hasTouch.current = ` away from `isValid` function so that `is` function doesn't have any side-effects

src/components/common/Interactive.tsx
```diff
@@ -108,36 +101,26 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
         left: keyCode === 39 ? 0.05 : keyCode === 37 ? -0.05 : 0,
         top: keyCode === 40 ? 0.05 : keyCode === 38 ? -0.05 : 0,
       });
-    },
-    [onKeyCallback]
-  );
-
-  const handleMoveEnd = useCallback(() => setDragging(false), []);
```
xnimorz, line 115: This was a source of potential errors and memory leaks.

When the component is about to be unmounted while the user is dragging the mouse:

1) We call `useIsomorphicLayoutEffect` cleanup
2) It checks that `isDragging` is true. 
3) `toggleDocumentEvents` won't be called
4) Component gets unmounted
=> we have a memory leak

Then the user releases  the mouse:
1) We call `handleMoveEnd`
2) It tries to `setState`
3) The component is unmounted => we receive the error

With the new flow we prevent memory leak and these errors

src/components/common/Interactive.tsx
```diff
@@ -108,36 +101,26 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
...
+    function toggleDocumentEvents(state: boolean) {
+      const touch = hasTouch.current;
       // add or remove additional pointer event listeners
-      const toggleEvent = state ? window.addEventListener : window.removeEventListener;
-      toggleEvent(hasTouched.current ? "touchmove" : "mousemove", handleMove);
-      toggleEvent(hasTouched.current ? "touchend" : "mouseup", handleMoveEnd);
-    },
-    [handleMove, handleMoveEnd]
-  );
+      const toggleEvent = state ? self.addEventListener : self.removeEventListener;
```
xnimorz, line 109: self is supported by all the browsers and it's smaller than window

src/components/common/Interactive.tsx
```diff
@@ -40,21 +43,31 @@ interface Props {
...
+  const hasTouch = useRef(false);
 
-  // Prevent mobile browsers from handling mouse events (conflicting with touch ones).
-  // If we detected a touch interaction before, we prefer reacting to touch events only.
-  const isValid = (event: MouseEvent | TouchEvent): boolean => {
-    if (hasTouched.current && !isTouch(event)) return false;
-    if (!hasTouched.current) hasTouched.current = isTouch(event);
-    return true;
-  };
+  const [handleMoveStart, handleKeyDown] = useMemo(() => {
```
xnimorz, line 50: Common `useMemo` prevents us from `useCallback` hell and it makes code easier to read, as all logic in one place

src/hooks/useEventCallback.ts
```diff
@@ -1,14 +1,12 @@
-import { useRef, useEffect, useCallback } from "react";
+import { useRef } from "react";
 
 // Saves incoming handler to the ref in order to avoid "useCallback hell"
-function useEventCallback<T>(handler?: (value: T) => void): (value: T) => void {
+export function useEventCallback<T>(handler?: (value: T) => void): (value: T) => void {
```
xnimorz, line 4: When the function is called it already created a ref to the function. Which means one of the provided function will be GCed. 
So there is no difference for us which function we will call, so we don't even need the `if` clause here as well as `useEffect`. 
It makes this custom hook much faster

xnimorz:
I'm not sure how to open the multitouch PR as it's based on the commit from this PR (you can see it here atm https://github.com/xnimorz/react-colorful/pull/1/files, I'll publish it to your repo after this diff)

src/components/common/Interactive.tsx
omgovich: Smart! 👌

src/components/common/Interactive.tsx
omgovich: The thing that bothers me now is abandoned event listeners.

I mean if we do not call `toggleDocumentEvents` on the component unmount the `touchmove` and `touchend` listeners are still active if the picker was unmounted in the middle of the dragging.

For example, removing `touchmove` and `touchend` events after a document unmounting requires an extra click on the screen.

src/hooks/useEventCallback.ts
omgovich: Looks great!

src/components/common/Interactive.tsx
xnimorz: In the previous implementation before the PR, we had the problem, that `toggleDocumentEvents` isn't called either. 
`isDragging` was set to true, so that the toggle cannot be executed. Which is the source of the memory leak. (after the component gets dismounted we are unable to change `isDragging` therefore we are unable to be unsubscribed)

The current solution doesn't call unsubscribe when the component dismounts, but as soon as the user generates a touch-up event we unsubscribe everything, plus all the handlers which were active don't make any effort, as we check containers (and they were false, as the component gets destroyed). 
If you think we want to unsubscribe at the time of components dismount, I can add useEffect with that. wdyt?

omgovich: Can't reproduce the memory leak and abandoned event listeners locally.

https://user-images.githubusercontent.com/206567/130655471-3d90f2d6-48d3-4434-ae45-ac5143cf2de7.mov

xnimorz: Oh, it's my mistake, I had the wrong test 🤦 Sorry for that 

Talking about cleaning up, I added commit which covers that case

src/components/common/Interactive.tsx
```diff
@@ -108,36 +101,27 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
...
 
-  useIsomorphicLayoutEffect(() => {
-    toggleDocumentEvents(isDragging);
-    return () => {
-      isDragging && toggleDocumentEvents(false);
-    };
-  }, [isDragging, toggleDocumentEvents]);
+    return [handleMoveStart, handleKeyDown, toggleDocumentEvents];
+  }, []);
+  useEffect(() => toggleDocumentEvents, []);
```
xnimorz, line 116: We don't need `useIsomorphicLayoutEffect` as `useEffect` won't be called on the server-side

Also, we don't need the check, as in the case we try to unsubscribe from the function, which doesn't exist as a subscribed browser will do nothing.

xnimorz:
Updated data: 
image

src/components/common/Interactive.tsx
omgovich: We use `useEffect` since Next.js shows tons of warnings about that =(

omgovich (approved):
Thanks a lot @xnimorz!

PR #156, opened 2021-09-12
Support multitouch

xnimorz:
Demo:

https://user-images.githubusercontent.com/1178825/132987284-850713f0-0cbd-4368-8cca-a256f7f2a00b.MP4

Solution:
Each touch event has its own id. 
As we don't use pointer-events, we may preserve the touch id, which makes it's possible to bind the touch and the definite bar. 

The only downside is that the size was slightly increased because of the `getPoint` fn:
image

### DIFF
--- a/src/components/common/Interactive.tsx
+++ b/src/components/common/Interactive.tsx
@@ -10,12 +10,25 @@ export interface Interaction {
 // Check if an event was triggered by touch
 const isTouch = (event: MouseEvent | TouchEvent): event is TouchEvent => "touches" in event;
 
+const getPoint = (touches: TouchList, touchId: null | number): Touch => {
+  for (let i = 0; i < touches.length; i++) {
+    if (touches[i].identifier === touchId) {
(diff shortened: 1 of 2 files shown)


src/components/common/Interactive.tsx
```diff
@@ -57,12 +71,15 @@ const InteractiveBase = ({ onMove, onKey, ...rest }: Props) => {
       preventDefaultMove(nativeEvent);
 
       if (isInvalid(nativeEvent, hasTouch.current) || !el) return;
-      hasTouch.current = isTouch(nativeEvent);
+      if (isTouch(nativeEvent)) {
+        hasTouch.current = true;
+        touchId.current = nativeEvent.changedTouches[0].identifier;
```
omgovich, line 76: Didn't know about `changedTouches` property 🤓
Hope it works everywhere we need.

omgovich:
Moved to #158

Issue #153, opened 2021-08-22
Iframe Support 


Chetan11-dev:
if rendered  react-colorful in iframe using https://www.npmjs.com/package/react-frame-component.   it does not work 
Could you guide me how to fix it. Maybe we will need to add eventListener to HTMLElement.defaultView for it instead of window.

omgovich:
Hi! Unfortunately, I don't have any experience with `react-frame-component` but I assume the reason for the problem is that CSS is injected into the parent window so the styles are not available inside of the iframe and the color picker seems invisible.

1. Could I ask you to prepare a small demo (for example in codesandbox.com) to make it easier for me to detect the problem.
2. As far as I know `defaultView` is a property of `document`, so it probably won't be a solution for the issue.

xnimorz:
Hey, I've checked the issue. The problem is that `react-colorful` is using `document` and `window` which `react-frame-component` provides separately as a context arguments: https://github.com/ryanseddon/react-frame-component/blob/master/src/Frame.jsx#L107 which is the pretty specific solution. 

The reason for this is that the react-frame-components renders react subtree as a portal and it doesn't execute the 
components code inside the `iframe` tab, it just moves the execution result into `iframe` node https://github.com/ryanseddon/react-frame-component/blob/master/src/Frame.jsx#L116-L117

In that case, it's better to have changes on `react-frame-component`, as tons of existed libraries won't support such a mode. 
The main reason for this is when you want to get current `document` you should have something like: `this.nodeRef.current.contentDocument`

Chetan11-dev:
Here is the reproduction https://github.com/Chetan11-dev/react-colorful-iframe. 
![Screenshot from 2021-08-30 18-41-18](https://user-images.githubusercontent.com/53407137/131344504-e85933de-cf86-4160-9f37-a578272eb8f4.png)
NO PROBLEM with css.(I can observe document head and add it to iframe head see reproduction.)
PROBLEM IS WITH eventlistners being attached 
SOLUTION: 
attach eventlistners to container's window. What I want you to do is to change code in src/components/common/Interactive.tsx like:
```js
  const toggleDocumentEvents = useCallback(
    (state) => {

      const containerOwnerWindow =(container.current as HTMLDivElement).ownerDocument.defaultView!
      // add or remove additional pointer event listeners
      const toggleEvent = state ? containerOwnerWindow.addEventListener : containerOwnerWindow.removeEventListener;
[...]

PR #166, opened 2021-11-14
use div to determine window

leoc4e:
Hi, I am facing some issues to when trying to use react-colorful inside an iframe.
- styles are injected into parent frame
- dragging could not change the input
- could not select full range of colors after the page is scrolled

These issue are similar to this [issus](https://github.com/omgovich/react-colorful/issues/153) that previously reported.

This PR change to use the `window` that rendered container div instead if possible for setting css or handling event.
It also handle the case that react-colorful used in multiple iframes or parent frame together


omgovich:
Thanks! Just came back from vacation. Going to review that as soon as I have time.

omgovich (approved):
Looks and works great ❤️ Thanks for your contribution!

Issue #152, opened 2021-08-10
About the 2d slider


diegohaz:
Currently, the saturation and brightness picker is using a single [`role="slider"`](https://www.w3.org/TR/wai-aria-1.2/#slider) element, which is supposed to be one-dimensional, but is being used to control a two-dimensional pane.

One of the problems with this approach is the lack of `aria-valuenow`, which is a required attribute on sliders, but can't be set here because it's dealing with two different axes.

This may be also confusing for screen reader users who expect the slider to work as described in the WAI-ARIA docs.

One of the solutions is using a [`role="grid"`](https://www.w3.org/TR/wai-aria-1.2/#grid) element. This has been discussed on https://github.com/w3c/aria/issues/432. The markup could be something like this:
```html
<div role="grid" aria-label="Saturation and Brightness" aria-rowcount="100">
  <!-- 80% saturation -->
  <div role="row" aria-label="Brightness" aria-rowindex="80">
    <!-- 20% brightness -->
    <div role="gridcell" aria-label="Saturation" aria-colcount="100" aria-colindex="20" />
  </div>
</div>
```

Another solution would be having two sliders:
```html
<div role="slider" aria-orientation="vertical" aria-valuenow="80" aria-label="Brightness" />
<div role="slider" aria-orientation="horizontal" aria-valuenow="20" aria-label="Saturation" />
```
Moving the arrow keys perpendicularly to the currently focused slider would move focus onto the other slider.

omgovich:
Hi guys! Thanks for using react-colorful and for the issue. I'll dig into the problem and try to roll out an approach that would provide better a11y and keep the bundle size as small as possible 👌

eGene:
Saturation and Hue components should accept props from user, this way we would be able to override aria-* attributes. This way I could just pass something like
```
saturationInputProp={{
  'aria-valuenow': '#12,34,56',
  'aria-label': 'Цвет'
}}
```
This also solves issue with hardcoded strings if I need non-english labels.

omgovich:
Hey guys! Don't think I forget about you. Just discussing the issue with guys who know more about a11y.

The thing that bothers me much and I still didn't get — which element must take the focus? `grid` or `gridcell`?

omgovich:
Hey @diegohaz. Could you please check https://omgovich.github.io/react-colorful/

I deployed a version with `role="grid"` (from the PR) there but seems to it that doesn't work well: My macOS VoiceOver doesn't say anything when I change value.

[...]

PR #177, opened 2022-02-24
Update ARIA attributes on ranges

### LINKED ISSUE #171: Slider ARIA properties
Really love this library, especially all the accessibility considerations built in! One thing I've noticed in implementing it though is an error from accessibility validators for a missing `aria-valuenow` property on both the Color and Hue sliders.

I had an explore of the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuetext) for the two properties `aria-valuetext` (currently used on these sliders) and `aria-valuenow`:

>The aria-valuetext attribute is used with the aria-valuenow attribute, not instead of it, unless that value is not known.
>
>aria-valuetext is only needed when the numeric value of aria-valuenow is not meaningful

The [WAI-ARIA docs for sliders](https://www.w3.org/TR/wai-aria-practices/#slider) also confirm `aria-valuenow` is required:

> Each slider element has the aria-valuenow property set to a decimal value representing the current value of the slider.

It seems that although `aria-valuetext` is what actually forms the description announced to screen reader users, the `aria-valuenow` is still required.

## Proposed change

In the case of the Hue component, the `valuetext` is already number, so I'm hoping it would be simple enough to change `aria-valuetext` to `aria-valuenow`, since the additional text description isn't needed.

It's less clear to me whether there's a corresponding numeric value that could be used for the Color slider - what do you think?

aitchiss:
Fixes https://github.com/omgovich/react-colorful/issues/171

## In the `Hue` component:

- `aria-valuetext` has been changed to `aria-valuenow`. Since the "human understandable" value is numeric, we don't need both `valuetext` and `valuenow`, and `valuenow` is the "required" attribute
- `aria-valuemin` has been set to 0 and `aria-valuemax` has been set to 360 to provide a bit more context for screen reader users

## In the `Alpha` component:

- `aria-valuenow` has been added. This is **in addition** to `aria-valuetext`. Screen readers will generally read out the more understandable `valuetext`, but `valuenow` is a required attribute to accompany it
- `aria-valuemin` has been set to 0 and `aria-valuemax` has been set to 100

## In the `Saturation` component:

No change has been made! I don't believe there is an available numeric representation of the current value, so in this case I believe we can omit `aria-valuenow` for the time being until https://github.com/omgovich/react-colorful/issues/152 is addressed.

Snapshot tests have been updated to reflect the new attributes 🙂 

Check out the [docs for `aria-valuenow`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-valuenow) for more context

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,7 +108,7 @@
       "path": "dist/index.module.js",
       "name": "RgbaStringColorPicker",
       "import": "{ RgbaStringColorPicker }",
-      "limit": "3 KB"
+      "limit": "3.1 KB"
     },
     {
       "path": "dist/index.module.js",
--- a/src/components/common/Alpha.tsx
+++ b/src/components/common/Alpha.tsx
@@ -35,6 +35,7 @@ export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
   };
 
   const nodeClassName = formatClassName(["react-colorful__alpha", className]);
+  const ariaValue = round(hsva.a * 100);
 
   return (
     <div className={nodeClassName}>
@@ -43,7 +44,10 @@ export const Alpha = ({ className, hsva, onChange }: Props): JSX.Element => {
         onMove={handleMove}
         onKey={handleKey}
         aria-label="Alpha"
-        aria-valuetext={`${round(hsva.a * 100)}%`}
+        aria-valuetext={`${ariaValue}%`}
+        aria-valuenow={ariaValue}
+        aria-valuemin="0"
+        aria-valuemax="100"
       >
         <Pointer
           className="react-colorful__alpha-pointer"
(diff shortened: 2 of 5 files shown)


omgovich:
Oh, wow! Thanks a lot for this contribution! I'll check it as far as I can!

omgovich:
Tested it locally with VoiveOver and added a simple test for that. Everything works well.
Thanks again!

aitchiss:
Thank you @omgovich!

smellai:
@aitchiss thanks for this improvement! 
@omgovich looks like the demo in the [homepage](https://omgovich.github.io/react-colorful/) is not updated yet with this change?

Looking forward the fix for #152 too

PR #181, opened 2022-06-05
fix attaching styles, check shadowroot

### LINKED ISSUE #180: react-colorful not working if we use it with react-shadow
here is a sample of the code, can you check it?

https://codesandbox.io/s/react-colorful-demo-forked-6vivtb?file=/src/App.js

definite2:
fixes #180

### DIFF
--- a/src/hooks/useStyleSheet.ts
+++ b/src/hooks/useStyleSheet.ts
@@ -1,30 +1,38 @@
+/* eslint-disable no-extra-boolean-cast */
 import { RefObject } from "react";
-
 import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect";
+import { getOwnerDocument } from "../utils/ownerDocument";
 import { getNonce } from "../utils/nonce";
 
 // Bundler is configured to load this as a processed minified CSS-string
 import styles from "../css/styles.css";
 
-const styleElementMap: Map<Document, HTMLStyleElement> = new Map();
+const styleElementMap: Map<Document | ShadowRoot, HTMLStyleElement> = new Map();
 
 /**
  * Injects CSS code into the document's <head>
  */
 export const useStyleSheet = (nodeRef: RefObject<HTMLDivElement>): void => {
   useIsomorphicLayoutEffect(() => {
-    const parentDocument = nodeRef.current ? nodeRef.current.ownerDocument : document;
-
-    if (typeof parentDocument !== "undefined" && !styleElementMap.has(parentDocument)) {
-      const styleElement = parentDocument.createElement("style");
+    const ownerDocument = getOwnerDocument(nodeRef.current);
+    const parentDocument = ownerDocument || document;
+    if (
+      parentDocument &&
+      typeof parentDocument !== "undefined" &&
+      !styleElementMap.has(parentDocument)
+    ) {
+      const styleElement = document.createElement("style");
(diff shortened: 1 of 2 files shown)


src/utils/shadowRoot.ts
```diff
@@ -0,0 +1,10 @@
+export const getShadowRoot = (node: any) => {
```
omgovich, line 1: Can we type it somehow? Would the code below work?

```suggestion
export const getShadowRoot = (node: HTMLElement) => {
```

src/utils/shadowRoot.ts
```diff
@@ -0,0 +1,10 @@
...
+  let parent = node && node.parentNode;
+  while (parent) {
+    if (parent.toString() === "[object ShadowRoot]") {
+      return parent;
+    }
+    parent = parent.parentNode;
+  }
```
omgovich, line 8: Is there any other option except stringifying a node? Doesn't seem very reliable)

I'm not a big expert in shadow root APIs, but I guess this might work `if (parent instanceof ShadowRoot)`

src/hooks/useStyleSheet.ts
```diff
@@ -1,30 +1,36 @@
...
 
 /**
  * Injects CSS code into the document's <head>
  */
 export const useStyleSheet = (nodeRef: RefObject<HTMLDivElement>): void => {
   useIsomorphicLayoutEffect(() => {
-    const parentDocument = nodeRef.current ? nodeRef.current.ownerDocument : document;
+    const parentDocument = nodeRef.current
+      ? !!getShadowRoot(nodeRef.current)
+        ? getShadowRoot(nodeRef.current)
```
omgovich, line 19: `getShadowRoot` runs while on nodes tree, so it's kinda expensive operation, so I would suggest running it just once (not 3 times) and putting result into a variable.

```js
const shadowRoot = getShadowRoot(nodeRef.current)
```

src/hooks/useStyleSheet.ts
```diff
@@ -1,30 +1,36 @@
...
+        : nodeRef.current.ownerDocument
+      : document;
 
-    if (typeof parentDocument !== "undefined" && !styleElementMap.has(parentDocument)) {
-      const styleElement = parentDocument.createElement("style");
+    if (parentDocument && !styleElementMap.has(parentDocument)) {
+      const styleElement = document.createElement("style");
```
omgovich, line 24: It would be `parentDocument` because changing that to `document` would break usage with libraries like https://www.npmjs.com/package/react-frame-component.

See https://github.com/omgovich/react-colorful/pull/166/files#diff-0e642e68af9db5d41afcf32c53546956ecbe896d844a354be630f80ab312e87a

omgovich:
Hi! Sorry for the delay: I'm going through a tough time in my life.

I left a few comments, could you take a look?

definite2:
> I'm going through a tough time in my life.

definite2:
> Hi! Sorry for the delay: I'm going through a tough time in my life.
> 
> I left a few comments, could you take a look?

Hi, @omgovich can you check again. I squash the update commit with the previous.

VitaliyR:
@omgovich hi, hopefully you doing alright, is there any intention to continue development of this lib or perhaps you consider to add somebody as a maintainer? 
Would be awesome to merge this PR

PR #182, opened 2022-06-05
Use `substring` instead of deprecated `substr`.

royeden:
Removes the `substr` method, as it's deprecated:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


omgovich:
Thanks, @royeden!

PR #189, opened 2022-08-03
Update package.json

AnotherHermit:
Add "types" to exports to handle issue where typescript wouldn't recognize exported types when using ESM.


omgovich:
Thanks!

